PR #11, opened 2017-04-12
Fixes a bug where `u2f` wasn't actually available to be called.

MicahZoltu:
This worked previously in the browser only if the end-developer took steps to ensure that `u2f` was added to the global namespace.  That isn't a particularly good pattern and it causes trouble with modern build systems that expect to be able to build a working dependency tree from the code.  This change simply makes it so the `u2f` object is available within the `ledger-comm-u2f` file as it expects and makes ledger work correctly with modern JS build tooling.

I left the file in the old location for backwards compatibility, though it should be removed at some point in the future (really, the whole browser folder should be removed IMO).

I also added an `isNode` check around the two require statements so the proper one is loaded based on environment.


MicahZoltu:
What is the release process like?  It doesn't appear releases auto-publish to NPM on merge to master.  Is there anything additional I need to do to get the library published to NPM or is it just whenever your internal process results in a release?

btchip:
I'll push it in a few days, wanted to merge in something else for the next release

PR #14, opened 2017-05-29
simple example for node

efx:
Figured this might be helpful for others trying out the node API.

### DIFF
--- a/examples/node-device-info.js
+++ b/examples/node-device-info.js
@@ -1,4 +1,4 @@
-const ledger = require('../src/index');
+const ledger = require('../src');
 ledger
 	.comm_node
 	.create_async()
--- a/examples/node-device-info.js
+++ b/examples/node-device-info.js
@@ -1,0 +1,10 @@
+const ledger = require('../src/index');
+ledger
+	.comm_node
+	.create_async()
+	.then(function(comm) {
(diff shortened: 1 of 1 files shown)


examples/node-device-info.js
```diff
@@ -0,0 +1,10 @@
+const ledger = require('../src/index');
```
antony: You can change this to simply:

```javascript
const ledger = require('../src');
```

examples/node-device-info.js
```diff
@@ -0,0 +1,10 @@
...
+ledger
+	.comm_node
+	.create_async()
+	.then(function(comm) {
+		console.log(comm.device.getDeviceInfo());
+	})
+	.catch(function(reason) {
```
antony: This project currently uses `Q` promises, so there is no `catch`:

```javascript
.fail(function(reason) {
...
}
```

efx: Q does actually support the standard `.catch`: https://github.com/kriskowal/q/wiki/API-Reference#promisecatchonrejected. I prefer keeping it this way because most Promise implementations support it as a standard.

antony: I wasn't aware of that - good to know :)

efx: no worries! thanks @antony

PR #16, opened 2017-07-20
Re-add Firefox Support

fabioberger:
This library currently does not work on Firefox. A regression was introduced in https://github.com/LedgerHQ/ledger-node-js-api/pull/11 where `ledger-comm-u2f` was modified to always use `u2f-api` instead of whatever `u2f` object was injected into the global scope. 

Firefox's U2F plugin would previously work with this library by injecting a `u2f` object onto the browser window object that was then picked up by `ledger-comm-u2f`. The `u2f-api` module it was replaced with only works on Chrome as it depends on the availability of a Chrome extension endpoint. 

This PR:
- Modifies `ledger-comm-u2f` so that it first checks for a `u2f` object in the global namespace. If none found, it defaults to using `u2f-api`.
- Renames `u2f-api` to `chrome-u2f-api` since a big reason this regression was introduced was the misunderstanding that `u2f-api` works with other browsers beyond Chrome.

@MicahZoltu, does this fix still work with your build system?


MicahZoltu:
Speaking _only_ about the fallback to `chrome-u2f-api` (the PR is pretty large and I didn't review it all) I suspect this won't cause a problem for us.  At the moment, I don't have things setup to try to build the project that uses this, but I don't see any reason off hand why this change would break things for me.

PR #18, opened 2017-09-05
No "q" dependency

maxx-t:
- Removed "q" dependency: the default "Promise" class is used.
- Turned all exports into factories in order to easily inject dependencies so that any "standard" promise library can be used (tested with "q" and "bluebird").
- A global factory is exposed with the legder object (it returns a ledger object that uses the promise library passed in parameter. For example: `ledger.build(require('bluebird'))`).
- The default ledger object uses the default "Promise" class (available under all modern browsers and nodejs).
- Added a .fail method to avoid breaking old implementations based on it.


btchip:
Thanks, I'll review that one quickly

gre:
Getting rid of Q library is definitely a good idea, now that Promise is available everywhere (if not, Promise polyfill can be used).

However, I think making all files returning a factory function creates some overhead for this library and I think we should stick with normal module code (and not injecting library in function arg).

I'm not sure why one would want to inject the Promise library (any reason you had in mind?) and this might lead to future bugs (imagine if some Promise library have subtle behavior difference).

Anyway, I feel it is unnecessary to allow to build the library with a specific promise lib because **Promise libraries are interoperable** by design (.then-able) so if you do something like `someQLib().then(()=>ledgerPromiseAPI())` it still returns a promise (of Q I think).

maxx-t:
Well, you're right. Actually I left the possibility to inject a promise library for browsers that didn't embedded one, but as this one only works (for now) with Firefox/Chrome it could indeed do without. Moreover, browsers without Promise functionality have probably more urgent to do than to implement the U2F API...

gre:
@maxx-t would you mind make changes on your PR to only have the part about migrating out from Q and to vanilla Promise code? thanks

maxx-t:
Done ;) I started over the current master in order to make it easier to review.

btchip:
Thanks for the work and review - as I'm currently modifying it for Segwit support, it looks like a great time to merge and test

PR #24, opened 2017-11-08
Upgrades node-hid.

MicahZoltu:
`node-hid 0.5.7` won't install from binaries under `node 8`.  This was fixed in `node-hid 0.5.7`.  I verified locally that _my_ stuff still works fine when `ledgerco` uses `node-hid 0.5.7`, and the author of that library seems to follow semver so I don't think there are any breaking changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "homepage": "https://github.com/LedgerHQ/ledger-node-js-api",
   "dependencies": {
-	  "node-hid": "0.5.4",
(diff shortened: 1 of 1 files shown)


MicahZoltu:
What is the status of a new NPM build?  Glancing at current NPM version and GitHub releases history, it appears the last release was on May 24th.

btchip:
I'm going to merge #26 in the coming days, then release as 1.2.0

PR #34, opened 2017-12-25
Changed buffer for varints

jimhansson:
If varint is of value bigger than 0xFFFF we need a buffer of size 5 not 4


gre:
thanks for raising the issue we have fixed it during the internal es6 refactoring :)

PR #36, opened 2017-12-30
Updates to Node HID 0.6.0.

MicahZoltu:
The primary change is addressing https://github.com/node-hid/node-hid/issues/226 which was making it really hard to use CI with Ledger Node JS API.


gre:
thanks for raising the issue we have fixed it during the internal es6 refactoring :) it's on master

PR #37, opened 2018-01-10
Split the projects into scoped libs + improve API doc

gre:
This is a major refactoring on the project to split the library into multiple libraries.

## new packages

We are introducing new libraries to replace the monolithic `ledgerco`.

to migrate, **you will need to depend on the relevant packages instead of `ledgerco`**. This will lighten your dependencies because `ledgerco` used to depend both on `node-hid` and `u2f-api`.

There is also new packages included (documented in the README.md and the API.md)

**packages:**

| Package                                                                  | Version                                                                                                                                       | Description                                                                                                                |
| ------------------------------------------------------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------- |
| [`@ledgerhq/hw-app-eth`](/packages/hw-app-eth)                           | [npm](https://www.npmjs.com/package/@ledgerhq/hw-app-eth)                           | Ethereum Application API                                                                                                   |
| [`@ledgerhq/hw-app-btc`](/packages/hw-app-btc)                           | [npm](https://www.npmjs.com/package/@ledgerhq/hw-app-btc)                           | Bitcoin Application API                                                                                                    |
| [`@ledgerhq/hw-transport-node-hid`](/packages/hw-transport-node-hid)     | [npm](https://www.npmjs.com/package/@ledgerhq/hw-transport-node-hid)     | Node implementation of the communication layer, using `node-hid` (USB)                                                     |
| [`@ledgerhq/hw-transport-u2f`](/packages/hw-transport-u2f)               | [npm](https://www.npmjs.com/package/@ledgerhq/hw-transport-u2f)               | Web implementation of the communication layer, using [U2F api](https://github.com/grantila/u2f-api)                        |
| [`@ledgerhq/hw-transport-http`](/packages/hw-transport-http)             | [npm](https://www.npmjs.com/package/@ledgerhq/hw-transport-http)             | communicate to the device over HTTP                                                                                        |
| [`@ledgerhq/hw-transport`](/packages/hw-transport)                       | [npm](https://www.npmjs.com/package/@ledgerhq/hw-transport)                       | The generic interface of the communication layer                                                                           |
| [`@ledgerhq/hw-http-proxy-devserver`](/packages/hw-http-proxy-devserver) | [npm](https://www.npmjs.com/package/@ledgerhq/hw-http-proxy-devserver) | DEVELOPMENT-only http server to communicate with the device with `hw-transport-node-hid` (bridge proxy for `hw-transport`) |

## api changes

There is also one more breaking change from the previous API: **the `_async` suffix has been REMOVED in all methods of the APIs**. The library is by nature asynchronous and it's common practice to do so in JS promise libraries. *We also highly recommend you to use async/await babel transformer to make your code much simpler.*

**BEFORE**

```js
const ledger = require("ledgerco");
function example() {
  ledger.comm_node.create_async().then(function(comm) {
    var btc = ledger.btc(comm);
    btc.getWalletPublicKey_async("44'/0'/0'/0").then(
       function(result) { console.log(result);}).catch(
       function(error) { console.log(error); });
  });
}
```

**AFTER**

```js
var Comm = require("@ledgerhq/hw-transport-node-hid"); // node
// OR
var Comm = require("@ledgerhq/hw-transport-u2f"); // web

var Btc = require("@ledgerhq/hw-app-btc");

function example() {
  Comm.list().then(devices => Comm.open(devices[0])).then(function(comm) {
    var btc = new Btc(comm);
    return btc.getWalletPublicKey("44'/0'/0'/0");
  }).then(function(result) {
    console.log(result);
  });
}
```

**AFTER (modern)**

```js
import Comm from "@ledgerhq/hw-transport-node-hid"; // node
// OR
import Comm from "@ledgerhq/hw-transport-u2f"; // web

import Btc from "@ledgerhq/hw-app-btc";

async function example() {
  const devices = await Comm.list();
  const comm = await Comm.open(devices[0]);
  const btc = new Btc(comm);
  const result = await btc.getWalletPublicKey("44'/0'/0'/0");
  console.log(result);
}
```

## `create` transport function is deprecated

Instead you should use `list` and `discover` and then use `open`.

## `listen` new transport function

We are also introducing a **`listen` function in Transport** implementations which is an alternative to list and allow to discover device as they come. It takes an observer and returns a subscription, as in this spec: https://github.com/tc39/proposal-observable . More examples will come later and implementation is yet to be fully implemented.

This is the first milestone for Ledger libs, we will have important other milestone soon! Cheers :)


gre:
remaining points I think to be fixed:

- the new discover() method. we need to make sure we actually are actively trying to discover new transports. e.g. for usb, we need to hook to new usb plugged. if a transport don't have such a possibility, we also need to .done() the observer.
  - for u2f, I don't think we can do that, not sure what we should do, not even sure we can know if there is a device available.. (if we can, we would do some polling to have discover working, would be neat feature)
- the "disconnect" event needs to be triggered on Transport. might not make sense for all transports.
- the list() method: I wonder if we should drop it, discover() is a more generic way to achieve this. maybe we would still keep list() for simplicity of usage tho?

new pkgs incoming:
- react-native hid transport
- react-native ble transport

PR #44, opened 2018-01-20
adds @ledgerhq/hw-app-xrp

hyperbolist:
* uses bip32-path
* doesn't rely on utils.js
* uses async/await
* includes tests
* generated signatures have been verified via outside tests

I understand the repo is going through a lot of changes right now, but I tried to make this conform to its modern vision based on what the sprinkled TODOs were asking for.


packages/hw-app-xrp/src/Xrp.js
```diff
@@ -0,0 +1,180 @@
...
+   * @param display optionally enable or not the display
+   * @param chainCode optionally enable or not the chainCode request
+   * @param ed25519 optionally enable or not the ed25519 curve (secp256k1 is default)
+   * @return an object with a publicKey, address and (optionally) chainCode
+   * @example
+   * const result = await xrp.getAddress("44'/144'/0'/0/0");
+   *
+   * {
+   *   "publicKey": "0324E5F600B52BB3D9246D49C4AB1722BA7F32B7A3E4F9F2B8A1A28B9118CC36C4",
+   *   "address": "rPwReL1dVYnVzFKShJhBxNPSbfX7rjJQEd"
```
gre: we might want to remove any "real" address from the code sample, just so people don't accidentally use them like it happened in the past and then contact us to reclaim them

probably you can just show a `const { publicKey, address } = result;`

packages/hw-app-xrp/README.md
```diff
@@ -0,0 +1,9 @@
+<img src="https://user-images.githubusercontent.com/211411/34776833-6f1ef4da-f618-11e7-8b13-f0697901d6a8.png" height="100" />
+
+## @ledgerhq/hw-app-eth
```
gre: please update the name of the lib. (that kinda sucks we will try to have template in the future)

packages/hw-app-xrp/src/Xrp.js
```diff
@@ -0,0 +1,180 @@
...
+   * sign a Ripple transaction with a given BIP 32 path
+   *
+   * @param path a path in BIP 32 format
+   * @param rawTxHex a raw transaction hex string
+   * @param ed25519 optionally enable or not the ed25519 curve (secp256k1 is default)
+   * @return a signature as hex string
+   * @example
+   * const signature = await xrp.signTransaction("44'/144'/0'/0/0", "12000022800000002400000002614000000001315D3468400000000000000C73210324E5F600B52BB3D9246D49C4AB1722BA7F32B7A3E4F9F2B8A1A28B9118CC36C48114F31B152151B6F42C1D61FE4139D34B424C8647D183142ECFC1831F6E979C6DA907E88B1CAD602DB59E2F");
+   *
+   * 30440220671dc22bed7006b2f170068ee0d9ddc5614b5fffdf438a69a956d9241c1d656b02207721ebf2d69c9c3938d6183bf087d10d786b9e94156ec46e4ddda9c8f7b95528
```
gre: same, probably that's ok not to show any result value. just in case someone use it

gre:
Thanks for the PR! here are some minor changes and it should be good.

> I understand the repo is going through a lot of changes right now, but I tried to make this conform to its modern vision based on what the sprinkled TODOs were asking for.

awesome work :) yes, the v2.0.0 of the libs we are about to merge should be stable enough to start developing these. I see you have used async/await and this is what we aim for in the future refactoring of existing Btc and Eth, so it's all good. in the future, we should provide even more helpers to simplify whatever factorization we can find between the different libraries (or using existing lib like you did for `bip32-path` dep. I had not time to refactor with this, but that's a good idea. it might probably make sense to refactor things from scratch for Btc and Eth in a near future).

regarding tests, we are looking for better solution in the future. final goal is to have headless simulators. another short term solution will be to have some "MockTransport" that record real devices input->output somewhere, and replay them on tests (like jest snapshot somehow). (this needs to be done on a test device where everyone have the same seed, AFAIK it should be deterministic)

hyperbolist:
Headless tests will be great.  I haven't looked at the blue-app-* projects in depth, but it would be really nice to test against them running in a some kind of debug CLI mode where you configure them with a known seed and use files for IO, but otherwise the apps would behave exactly as they do when they're running on the hardware.  Might be more trouble than it's worth, or impossible.  But the confidence would be nice.

These tests merely verify the hardware communication is successful.  I created a separate test for myself to verify that the signatures I generated were valid by using ripple-lib on the test net.  I imagine everyone will want to do something similar when they're developing a new hw-app-*, or even blue-app-*.  Maybe there's room for a test project that submits test transactions to each respective test net?  Again, this might be more trouble than it's worth.  Each coin is so different in how they distribute test net credentials, so it might not be possible to create a one-size-fits-all project to run test net transactions in an automated way.  But it's something I'm thinking about.

gre:
yeah I think there are lot of rooms to improve tests and is warmly welcomed to contribution :) so is the documentation that we'll try to get improved in the next weeks.

as you mentioned, it can be pretty tricky to test everything, like for instance we can't mock the entire blockchain of each coin^^ but at least we can check that a given output is expected if you give a given input. if this output is hardcoded and has been double check it is correct, we have some level of confidence. I don't know if you are familiar to that: https://facebook.github.io/jest/docs/en/snapshot-testing.html but it's something interesting because very quick to use. the risk is to have false-positive / incorrect expected output.



Thanks for the PR

PR #58, opened 2018-02-01
Feature/segwit

amougel:
Add segwit support for creating payment transaction

### DIFF
--- a/packages/hw-app-btc/src/Btc.js
+++ b/packages/hw-app-btc/src/Btc.js
@@ -716,13 +716,18 @@ btc.signP2SHTransaction(
    * @example
 const tx1 = btc.splitTransaction("01000000014ea60aeac5252c14291d428915bd7ccd1bfc4af009f4d4dc57ae597ed0420b71010000008a47304402201f36a12c240dbf9e566bc04321050b1984cd6eaf6caee8f02bb0bfec08e3354b022012ee2aeadcbbfd1e92959f57c15c1c6debb757b798451b104665aa3010569b49014104090b15bde569386734abf2a2b99f9ca6a50656627e77de663ca7325702769986cf26cc9dd7fdea0af432c8e2becc867c932e1b9dd742f2a108997c2252e2bdebffffffff0281b72e00000000001976a91472a5d75c8d2d0565b656a5232703b167d50d5a2b88aca0860100000000001976a9144533f5fb9b4817f713c48f0bfe96b9f50c476c9b88ac00000000");
    */
-  splitTransaction(transactionHex: string): Transaction {
+  splitTransaction(transactionHex: string, isSegwitSupported: boolean): Transaction {
     const inputs = [];
     const outputs = [];
+    var witness = false;
     let offset = 0;
     const transaction = Buffer.from(transactionHex, "hex");
     const version = transaction.slice(offset, offset + 4);
     offset += 4;
+    if (isSegwitSupported && ((transaction[offset] == 0) && (transaction[offset + 1] != 0))) {
+        offset += 2;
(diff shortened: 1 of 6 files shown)


packages/hw-app-btc/src/Btc.js
```diff
@@ -166,6 +181,32 @@ export default class Btc {
...
+    sha.update(this.serializeTransaction(transaction, true));
+    let hash = sha.digest();
+    sha = createHash("sha256");
+    sha.update(hash);
+    hash = sha.digest();
+    const look1 = indexLookup & 0xff;
+    const look2 = (indexLookup >> 8) & 0xff;
+    const look3 = (indexLookup >> 16) & 0xff;
+    const look4 = (indexLookup >> 24) & 0xff;
+    let data = Buffer.from([look1, look2, look3, look4]);
```
gre: not very important but I would suggest to use `data = Buffer.alloc(4);` and then do `data.writeUInt32LE(indexLookup, 0)` . it makes it a bit less error prone and simple. 

( ultimately in the future, i wish we can find an interesting abstraction so we don't directly deal with buffer but with a descriptive specification where we express the binary format )

packages/hw-app-btc/src/Btc.js
```diff
@@ -493,55 +544,82 @@ btc.createPaymentTransactionNew(
...
           targetTransaction.inputs[i].script =
             inputs[i].length >= 3 && typeof inputs[i][2] === "string"
               ? Buffer.from(inputs[i][2], "hex")
-              : regularOutputs[i].script;
+              : !segwit
+                ? regularOutputs[i].script
+                : Buffer.concat([
+                    Buffer.from("76a914", "hex"),
+                    this.hashPublicKey(publicKeys[i]),
+                    Buffer.from("88ac", "hex")
```
gre: what are these constants? we could put them statically just so we actually have them named?

packages/hw-app-btc/src/Btc.js
```diff
@@ -493,55 +544,82 @@ btc.createPaymentTransactionNew(
...
+                  ]);
           return this.startUntrustedHashTransactionInput(
-            firstRun,
+            !segwit ? firstRun : false,
```
gre: `!segwit && firstRun`

packages/hw-app-btc/src/Btc.js
```diff
@@ -551,12 +629,28 @@ btc.createPaymentTransactionNew(
...
-          this.serializeTransaction(targetTransaction),
-          outputScript,
-          lockTimeBuffer
+        var result = Buffer.concat([
+          this.serializeTransaction(targetTransaction, false),
+          outputScript
         ]);
 
+        if (segwit) {
+          var witness = Buffer.from("", "hex");
```
gre: can just do `Buffer.alloc(0)` instead

packages/hw-app-btc/src/Btc.js
```diff
@@ -705,13 +800,24 @@ btc.signP2SHTransaction(
...
     const inputs = [];
     const outputs = [];
+    var witness = false;
     let offset = 0;
     const transaction = Buffer.from(transactionHex, "hex");
     const version = transaction.slice(offset, offset + 4);
     offset += 4;
+    if (
+      isSegwitSupported &&
+      (transaction[offset] == 0 && transaction[offset + 1] != 0)
```
gre: prefer === and !==

packages/hw-app-btc/src/Btc.js
amougel: These are operators for the spending script. I' ll have them named

PR #63, opened 2018-02-05
Feature/timestamp

amougel:
Adding timestamp management to prepare for future altcoins integration

### DIFF
--- a/packages/hw-app-btc/src/Btc.js
+++ b/packages/hw-app-btc/src/Btc.js
@@ -907,7 +907,11 @@ const outputScript = btc.serializeTransactionOutputs(tx1).toString('hex');
 
   /**
    */
-  serializeTransaction(transaction: Transaction, skipWitness: boolean) {
+  serializeTransaction(
+    transaction: Transaction,
+    skipWitness: boolean,
+    timestamp?: Buffer
+  ) {
     let inputBuffer = Buffer.alloc(0);
     let useWitness =
       typeof transaction["witness"] != "undefined" && !skipWitness;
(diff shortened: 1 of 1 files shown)


packages/hw-app-btc/src/Btc.js
```diff
@@ -32,6 +32,20 @@ export default class Btc {
...
+  };
+
+  toHexInt = number => {
+    return (
+      this.toHexDigit((number >> 24) & 0xff) +
+      this.toHexDigit((number >> 16) & 0xff) +
+      this.toHexDigit((number >> 8) & 0xff) +
+      this.toHexDigit(number & 0xff)
+    );
+  };
```
gre: can these 2 be implemented using Buffer instead? 

something like `buffer=Buffer.alloc(4); buffer.writeUInt32LE(number, 0); return buffer.toString("hex")` (not tested) ?

packages/hw-app-btc/src/Btc.js
```diff
@@ -497,6 +515,19 @@ btc.createPaymentTransactionNew(
...
+    if (hasTimestamp) {
+      hasTimestamp().then(
+        date =>
+          (timestamp = Buffer.from(
+            this.toHexInt(date / 1000)
+              .match(/([0-9a-f]{2})/g)
+              .reverse()
+              .join(""),
+            "hex"
+          ))
```
gre: I think it's better if we can directly create the Buffer with the number. without using a temporary string conversion

packages/hw-app-btc/src/Btc.js
amougel: Yes, I was just transcribing the code (not trying to improve it right away). I'll fix it.

PR #64, opened 2018-02-06
add 'currencies' library (forked out of common)

gre:


packages/currencies/src/index.js
```diff
@@ -0,0 +1,35 @@
...
+  getCurrencyByCoinType,
+  getCurrencyByCoinTypeSafe,
+  countervalueForRate,
+  encodeURIScheme,
+  decodeURIScheme
+};
+
+import type { Currency, Unit, Rate, UnitValue } from "./types";
+
+export type { Currency, Unit, Rate, UnitValue };
```
meriadec: not sure for flow types, but doesn't this works?

```
export type { Currency, Unit, Rate, UnitValue } from './types'
```

PR #66, opened 2018-02-08
add create-dapp and web3-subprovider

gre:
yes


packages/create-dapp/template/README.md
```diff
@@ -0,0 +1,40 @@
...
+```
+cd src/dapp
+truffle compile
+truffle deploy --network development
+```
+
+Run the web app:
+
+```
+yarn start
```
loeck: Before, you ask to use `npm` and now `yarn` maybe just one or the other?

gre:
BTW the web3-subprovider was inspired by https://github.com/Neufund/ledger-wallet-provider/ from @banciur so kudos to Neufund team.
Code was completely refactored, some features added and tiny bugfixes (like the networkId & 0xff thing).

banciur:
@gre good stuff. I should mention @LogvinovLeon who initially worked on it 

Anyway one thing that I forgot to fix in our doc. To run `web3-provider-engine` under node you cannot use `RpcSubprovider` as it doesn't work in node. You have to switch to `FetchSubprovider` - it also works in browsers.

gre:
@banciur oh interesting! thanks

gre:
@banciur I have not anticipated that web3-provider could be used in Node context, I thought it was meant for web only, what is the usecase to run it in node? if there is a need, we might need to change `@ledgerhq/web3-subprovider` lib so it acceptes a transport, instead of having a hardcoded dependency on u2f

gre:
@banciur I have not anticipated that web3-provider could be used in Node context, I thought it was meant for web only, what is the usecase to run it in node? if there is a need, we might need to change `@ledgerhq/web3-subprovider` lib so it acceptes a transport, instead of having a hardcoded dependency on u2f (actually it might be a better idea anyway, as later we will have transport over web-bluetooth, web-usb. I mean in 2018,2019,2020, these spec will eventually settle^^)

Issue #65, opened 2018-02-07
inconsistent unit for timeout


gre:
u2f use seconds: https://github.com/LedgerHQ/ledgerjs/blob/master/packages/hw-transport-u2f/src/TransportU2F.js#L71
rest uses milliseconds: https://github.com/LedgerHQ/ledgerjs/blob/master/packages/hw-transport/src/Transport.js#L241

decide which to keep. I think milliseconds is more conventional in JavaScript. (e.g. setTimeout, setInterval are in ms)

PR #67, opened 2018-02-08
add Transport#setExchangeTimeout and use it in U2F instead of open timeout

gre:
this was a mistake to use the open timeout because the exchange timeout is another concept.
moreover, this was a bug that the TransportU2F was expecting a timeout in seconds and not in milliseconds like all other parts. Fixes https://github.com/LedgerHQ/ledgerjs/issues/65

Technically, this commit is a breaking change because if you were using the U2F timeout, you should now use the .setExchangeTimeout instead. otherwise you will have the default of 30s. We probably will unroll a major release to be safe.

NB: some transport does not implement openTimeout / exchangeTimeout at the moment. we will incrementally improve this.


packages/hw-transport-u2f/src/TransportU2F.js
```diff
@@ -65,19 +65,13 @@ export default class TransportU2F extends Transport<null> {
...
-    super();
-    this.timeoutSeconds = timeoutSeconds;
-  }
-
   /**
    * static function to create a new Transport from a connected Ledger device discoverable via U2F (browser support)
    */
-  static open(_: *, timeout?: number): Promise<TransportU2F> {
-    return Promise.resolve(new TransportU2F(timeout));
+  static open(): Promise<TransportU2F> {
```
amougel: Shouldn't we keep a timeout for opening?

PR #74, opened 2018-02-10
Stellar App API

lenondupe:


lenondupe:
Not sure why Travis is still failing. It doesn't give any useful indication what might be wrong still. It just says it's stuck. Potentially, if you it's running the actual tests on a device the Stellar app needs to be installed.

gre:
Hi @lenondupe , thanks for this PR we should review it in the week. yeah, Travis has been painful this past week as our project grows, it tends to timeout.

gre:
Travis timing out on this is a mystery to me

gre:
@lenondupe hi :) ok so it looks like the last run did a bit better. I wonder if the fact you added the missing dependency in test helped (that would mean one of our command tends to hang if there are missing pkgs? 🤔). Anyway, looking at the latest test, it looks like you may need to run prettier over the package? `yarn run prettier` should do the work

lenondupe:
Yes I think that did the job. Now also committed the prettier code.

gre:
ok so this was a problem with `flow` when running it in parallel, i have forced the concurrency to be 1. so it run sequentially

lenondupe:
Thanks @gre for quick review and merge!

gre:
@lenondupe I also tested your test and works fine. I have included in the work around "test snapshotting" so we have regression test now! https://github.com/LedgerHQ/ledgerjs/pull/81

lenondupe:
Oh that's really super!

Issue #79, opened 2018-02-13
improve error messages


gre:
this is a continuous task but basically instead of `Invalid Status 9f5a` we should have something like `Invalid Status 9f5a: Please check you are running on the correct app` (trying to have the most precise error message, yet visually printable), like on MEW.

PR #91, opened 2018-02-22
Fixes https://github.com/LedgerHQ/ledgerjs/issues/79

gre:
Instead of `Invalid status 6985` thrown for instance when refusing to sign a transaction, you will now gets: `Ledger device: Condition of use not satisfied (denied by the user?) (0x6985)` which is less crypto error message. Idea is ultimately, all the error.message are printable for the user. note that the message is printable but the error object is also exploitable programmatically because we have an .id field on it, as well as complementary field (like .statusCode, .statusText )

and also makes the StatusCodes constants consistant with code found on https://github.com/LedgerHQ/blue-app-btc/blob/d8a03d10f77ca5ef8b22a5d062678eef788b824a/include/btchip_apdu_constants.h#L85-L115

### DIFF
--- a/packages/hw-transport/src/Transport.js
+++ b/packages/hw-transport/src/Transport.js
@@ -27,36 +27,61 @@ export type Observer<Ev> = {
 
 /**
  * all possible status codes.
- * @see https://ledgerhq.github.io/btchip-doc/bitcoin-technical.html#_status_words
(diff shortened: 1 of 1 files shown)


gre:
probably safer to have this merged when we do the next major. lil' breaking change of renaming the status code constants

PR #93, opened 2018-02-23
localize the formatCurrencyUnit

gre:
you can provide a locale option


gre:
i shouldn't have touched the one on common, fixed ;)

PR #99, opened 2018-03-01
Detect attempt at signing multi-op Stellar transaction and throw descriptive error

lenondupe:
Stellar allows to batch multiple operations in a single transaction but the Ledger Stellar app does not support this. This patch detects this error condition and provides a user-friendly error message.

### DIFF
--- a/packages/hw-app-str/package.json
+++ b/packages/hw-app-str/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ledgerhq/hw-app-str",
-  "version": "4.2.0",
+  "version": "4.2.1",
   "description": "Ledger Hardware Wallet Stellar Application API",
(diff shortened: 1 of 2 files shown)


packages/hw-app-str/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ledgerhq/hw-app-str",
-  "version": "4.2.0",
+  "version": "4.2.1",
```
gre: would you mind not changing the version? we are doing the versioning globally at the moment using lerna and our tooling will automatically update the versions

lenondupe: no problem, I didn't exactly know the scheme anyway. I've changed it back.

PR #116, opened 2018-03-15
Automatic building of React and React Native SVG icons

MortalKastor:
+ New currencies icons


packages/currencies/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ledgerhq/currencies",
-  "version": "4.5.0",
+  "version": "4.6.0",
```
gre: we don't bump the version in PRs , it is done automatically by lerna when we are publishing from master

packages/currencies/src/reactNative.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+import * as icons from "./data/icons/reactNative";
+
+const iconsByCoinType = {
+  "0": icons.bitcoin,
+  "1": icons.bitcoin,
+  "2": icons.litecoin,
+  "3": icons.dogecoin,
+  "5": icons.dash
```
gre: we should unify some logic between both react and reactNative, basically should do:

coinTypeToIconName.js
```js
export default {
  "0": "bitcoin",
  "1": "bitcoin",
  "2": "litecoin",
}
```

and then on each react & reactNative.js

```js
export function getIconByCoinType(coinType: number): ?Icon {
  return icons[coinTypeToIconName[coinType]];
}
```

packages/currencies/package.json
```diff
@@ -34,7 +34,9 @@
...
-    "build": "node scripts/buildReactIcons.js && cd ../.. && export PATH=$(yarn bin):$PATH && cd - && babel --source-maps -d lib src && flow-copy-source -v src lib",
-    "watch": "node scripts/buildReactIcons.js && cd ../.. && export PATH=$(yarn bin):$PATH && cd - && babel --watch --source-maps -d lib src & flow-copy-source -w -v src lib"
+    "build":
+      "node scripts/buildReactIcons.js && cd ../.. && export PATH=$(yarn bin):$PATH && cd - && babel --source-maps -d lib src && flow-copy-source -v src lib",
+    "watch":
+      "node scripts/buildReactIcons.js && cd ../.. && export PATH=$(yarn bin):$PATH && cd - && babel --watch --source-maps -d lib src & flow-copy-source -w -v src lib"
   }
-}
\ No newline at end of file
+}
```
meriadec: you should ignore package.json from prettier, as this formatting will be overwritten everytime you yarn add something

Issue #112, opened 2018-03-10
Ledger device unreachable. Make sure the device is…unlocked and with the correct application opened.


mandelmonkey:
Hi
I am using the latest btc app ver 1.2.3 and nano s firmware 1.4.1 chrome Version 64.0.3282.186 (Official Build) (64-bit) on Mac OS

I am trying to get the address on the browser using the following code but I get the error

```Ledger device unreachable. Make sure the device is…unlocked and with the correct application opened.```

previously browser mode had to be set however in the new app the option is not available (I heard it was not needed anymore)

Cable seems to be ok, I can use the chrome btc app no problem

```
    if (TransportU2F.isSupported()) {

      TransportU2F.create().then(function(transport) {

        const btc = new Btc(transport);

        btc.getWalletPublicKey("44'/0'/0'/0").then(function(address) {

          console.log(address);

        }).catch(function(error) {

          console.log(error);

        });
      }).catch(function(error) {

        console.log(error);

      });
    }
```

gre:
@mandelmonkey yes, we are aware of the problem, there currently is an issue exclusively affecting communication of U2F with the BTC app on the latest firmware. We are working on the fix and a new btc app should get release ASAP. Let's leave the issue opened and we'll get back to you when it's fixed

mandelmonkey:
awesome! just wanted to check if I was doing something wrong, is it an issue on the btc app side? if so is there a branch of the btc I could use?

gre:
well, is it an issue on the hardware app (not having u2f working properly), not on the JS side. so nothing you can easily do for now

mandelmonkey:
so its an issue on the blue-btc-app? I am using my own fork of the app so if there  is a branch or pull request for a fix would be handy to see, if not I can wait for the master branch to be updated :-)

gre:
oh neat! no idea where and what the bug is, that's a bit beyond my expertise, as it's using the new firmware SDK I'm not even sure a previous version of the app would work 🤔 (maybe?). I guess better wait the fix to come, unless it's urgent for you & you want to figure it out, but otherwise they are working on it and it probably should get released next week.

roosmaa:
I did some digging into this, wrote about it in the Slack #ledgerjs channel, but here's a copy-paste of it:

> so first of all, I think why ETH still works is because the ETH app (at least the repo) hasn't been migrated to the new U2F library bundled in the SDK 1.4.1
> 
[...]

PR #123, opened 2018-03-21
Disabling device pinging in U2f

amougel:
Allow usage of post 1.4.1 firmware apps with U2F by temporarily disabling device pinging on opening new U2F Transport


gre:
ctx #112

PR #125, opened 2018-03-23
web3-subprovider: fix `address unknown error` due to address capitalization

williamchong007:
in `"web3": "^1.0.0-beta.28"`, when calling `web3.eth.personal.sign`, the payload address will be casted to lower case by input address formatter
https://github.com/ethereum/web3.js/blob/1.0/packages/web3-core-helpers/src/formatters.js#L396

however the `addressToPathMap` is storing the `capitalization checksum address` as key, thus the lower case address mismatch and the sign fail.

This PR fixes this by casting all input addresses and `addressToPathMap` keys to lower case


williamchong007:
@gre any schedule will this be released? basically can't use web3 1.0 for signing in the current version

gre:
@williamchong007 v4.7.3 should have that

PR #127, opened 2018-03-28
currencies: add chopCurrencyUnitDecimals

gre:
typically will be used for the input text because we don't want to aggressively re-format it on key stroke

### DIFF
--- a/packages/currencies/tests/index.test.js
+++ b/packages/currencies/tests/index.test.js
@@ -10,6 +10,7 @@ import {
   countervalueForRate,
   formatCurrencyUnit,
   parseCurrencyUnit,
+  chopCurrencyUnitDecimals,
   formatShort,
   decodeURIScheme,
   encodeURIScheme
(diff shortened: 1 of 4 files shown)


packages/currencies/tests/index.test.js
```diff
@@ -169,6 +170,20 @@ test("formatShort", () => {
...
 });
 
+test("chopCurrencyUnitDecimals", () => {
+  expect(chopCurrencyUnitDecimals(getFiatUnit("EUR"), "1")).toBe("1");
+  expect(chopCurrencyUnitDecimals(getFiatUnit("EUR"), "1,234")).toBe("1,234");
+  expect(chopCurrencyUnitDecimals(getFiatUnit("EUR"), "1,234.56")).toBe(
+    "1,234.56"
+  );
+  expect(chopCurrencyUnitDecimals(getFiatUnit("EUR"), "1,234.5678")).toBe(
+    "1,234.56"
```
meriadec: shouldn't it be `1,234.57` ?

gre: no. the method is called chop, not round ;)

it's meant for input text right? If I type `12.34` and then i add a `9` , it should just do nothing, not switch my input to `12.35`. and then `12.36` again if I type 9

meriadec: make sense :ok_hand: :

PR #132, opened 2018-03-30
currencies: changes in cryptocurrencies & fiat type

gre:
- fiat has been a bit reduced to only contain the fiat that have pairs in crypto world ( according to https://min-api.cryptocompare.com/data/all/exchanges )
- fiat now also include a "ticker" field. which is the same as "code" at the moment. but that we should use to refer a fiat in API / exchanges etc
- currency also have a "ticker" field for the same usecase. it was a hack to use currency.units[0].code and we'll migrate these
- fiat unit have its own type

### DIFF
--- a/packages/currencies/src/data/fiat.js
+++ b/packages/currencies/src/data/fiat.js
@@ -1,216 +1,69 @@
 //@flow
-import type { Unit } from "../types";
+import type { FiatUnit } from "../types";
 // inspired by https://github.com/smirzaei/currency-formatter/blob/master/currencies.json
-const units: { [key: string]: Unit } = {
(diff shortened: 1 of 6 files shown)


packages/currencies/src/data/fiat.js
```diff
@@ -1,39 +1,13 @@
 //@flow
-import type { Unit } from "../types";
+import type { FiatUnit } from "../types";
 // inspired by https://github.com/smirzaei/currency-formatter/blob/master/currencies.json
-const units: { [key: string]: Unit } = {
+const units: { [key: string]: Object } = {
```
meriadec: Can't we use `FiatUnit` instead of `Object`?

gre: only if we put the `ticker` inline. maybe we can

PR #150, opened 2018-05-11
Load accounts in a more efficient way

gbalabasquer:
Not need to go to the device for each address to load

### DIFF
--- a/packages/web3-subprovider/package-lock.json
+++ b/packages/web3-subprovider/package-lock.json
@@ -1,0 +1,235 @@
+{
+  "name": "@ledgerhq/web3-subprovider",
+  "version": "4.11.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "bindings": {
+      "version": "1.3.0",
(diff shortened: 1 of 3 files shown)


gre:
Hi @gbalabasquer , so this will derive the addresses on JS side instead of on the device, smart 🙂
The problem i'm having is the dependency `hdkey` is bringing too much dependency. In context of web3 , it might be ok because it's a lot of underlying lib, but still I think this should be somehow optional.
especially `hdkey` depends on `crypto` lib, which is usually painful to have it correctly polyfilled. Especially for instance in context of React Native.

I'm not exactly sure what best choice we have here to satisfy this feature, like maybe this could be a different library, or also it could be a different entry point.

Like we could make an optionalDependency & have a different entry point like

```
import createLedgerSubprovider from "@ledgerhq/web3-subprovider/lib/hdkey";
```

gre:
another way could be to abstract out an option that allows you to give the derivation implementation.
the `"@ledgerhq/web3-subprovider/lib/hdkey"` would just use that to provide the extra logic

e.g.

```
export default (getTransport, opts) => createLedgerSubprovider(getTransport, { ...opts, derivateFromXPUB: xpub => ... })
```

?

PR #171, opened 2018-06-29
feat(hw-app-str): generalize for more networks

Akuukis:
At [ledger-app-stellar #11](https://github.com/lenondupe/ledger-app-stellar/issues/11) it was clear that each Stellar technology based network will have to make it's own app. Nevertheless, I see that with this change it's easy to use the same companion Javascript library for all of them.

Tested and works for me with [FIC Network](https://github.com/ficnetwork/ledger-app-fic).

Is this the right way to move forward?

### DIFF
--- a/packages/hw-app-str/src/utils.js
+++ b/packages/hw-app-str/src/utils.js
@@ -104,7 +104,6 @@ export function hash(data: Buffer) {
 }
 
 export function checkStellarBip32Path(path: string): void {
-
   path.split("/").forEach(function(element) {
     if (!element.toString().endsWith("'")) {
       throw new Error(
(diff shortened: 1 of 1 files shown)


Akuukis:
Updated the PR to latest master. Stellar contributors @lenondupe @phra , what are your thoughts on this?

lenondupe:
I'd prefer to simply remove the check altogether to be honest. If there is there a dedicated status word that is returned for when a request is made that does not correspond to the lock path of the app we may catch it and make a more descriptive message for the caller about what's going but otherwise I'd simply just remove the check.

Akuukis:
@lenondupe Ok, makes sense. I've done so, can you please review?

Maintainers, I am confused why Travis is failing, and why it's out-of-date. Travis doesn't have any errors, nor I see any new changes in master.

Akuukis:
@lenondupe bump.

lenondupe:
Sorry, yeah, looks good to me. @gre can this be merged?

PR #174, opened 2018-07-02
Fixes the derivation path.

MicahZoltu:
The path in the readme was not a valid BIP43/BIP44 path.  This fixes it to be BIP44 compliant.


gre:
you are totally right

PR #190, opened 2018-07-23
fix "TypeError: sub is undefined"

levity:
fix a simple Javascript scoping error, so that we don't get "TypeError: sub is not defined" when timeout occurs (e.g. because Ledger is not connected).

### DIFF
--- a/packages/hw-transport/src/Transport.js
+++ b/packages/hw-transport/src/Transport.js
@@ -288,12 +288,6 @@ TransportFoo.create().then(transport => ...)
   ): Promise<Transport<Descriptor>> {
     return new Promise((resolve, reject) => {
       let found = false;
-      const listenTimeoutId = setTimeout(() => {
-        sub.unsubscribe();
-        reject(
-          new TransportError(this.ErrorMessage_ListenTimeout, "ListenTimeout")
-        );
-      }, listenTimeout);
       const sub = this.listen({
         next: e => {
           found = true;
@@ -317,6 +311,12 @@ TransportFoo.create().then(transport => ...)
           }
         }
       });
+      const listenTimeoutId = setTimeout(() => {
+        sub.unsubscribe();
+        reject(
+          new TransportError(this.ErrorMessage_ListenTimeout, "ListenTimeout")
+        );
+      }, listenTimeout);
     });
   }
 


gre:
neat! thanks a lot.

related https://github.com/MyCryptoHQ/MyCrypto/issues/2064

gre:
are you sure this is actually fixing the bug? I don't really understand how this was a JavaScript scoping problem.
defining a variable after it's used asynchronously is supposed to work. Flowtype didn't even see the problem 🤔  

This code in Node runs fine.

```
setTimeout(() => {
  console.log(a.toUpperCase());
}, 100);
const a = "toto";
```

gre:
I think the problem is somewhere else. I actually suspect the code `sub.unsubscribe();` that runs in the `next` should actually be `if (sub) sub.unsubscribe()`. because i'm not sure it's guarantee this next() runs asynchronously..
I think that's the actual bug fix. the current PR is only esthetic but i'm pretty sure it does not fix the bug.

levity:
Perhaps the issue only shows up after the code is transpiled by Babel? All I know is that I was getting the error and moving the code as shown in this PR fixed the issue.

gre:
@levity  but how can you tell that `TypeError: sub is undefined` was coming from the one in setTimeout? there are two occurences of `sub.unsubscribe()` and looking at the code, i'm pretty sure that the bug comes from the other one.

levity:
The stack trace for the error in the browser's console points to the one in setTimeout.

levity:
I think it's due to Babel. This is how the code looks after transpilation (I added some comments):

```js
function(r, i) { // this is the argument to new Promise
    var a = !1
      , s = setTimeout(function() { // this is `const listenTimeoutId`
        u.unsubscribe(),
        i(new o(t.ErrorMessage_ListenTimeout,"ListenTimeout"))
    }, n)
      , u = t.listen({ // this is `const sub`
        next: function(n) {
            a = !0,
            u.unsubscribe(),
            clearTimeout(s),
            t.open(n.descriptor, e).then(r, i)
        },
        error: function(t) {
            clearTimeout(s),
            i(t)
        },
        complete: function() {
            clearTimeout(s),
            a || i(new o(t.ErrorMessage_NoDeviceFound,"NoDeviceFound"))
        }
    })
}
```

levity:
Well, maybe not -- that still works if I plug in values to make it a standalone example. Is it possible that `listen` just returns undefined under certain conditions?

gre:
no I think the problem is that next is called somewhere in sync and not asynchronously.

so basically if listen() calls `o.next()` straightaway before returning the `{unsubscribe}` , obviously `sub` is undefined at this time

PR #191, opened 2018-07-24
Only poll USB devices list when a change is detected

MortalKastor:


packages/hw-transport-node-hid/src/TransportNodeHid.js
```diff
@@ -68,6 +69,10 @@ export default class TransportNodeHid extends Transport<string> {
     listenDevicesPollingSkip = conditionToSkip;
   };
 
+  static setListenDevicesDebugMode = (bool: boolean) => {
```
gre: it would be better if we would allow to set a log function instead. I moved to this paradigm in ledgerjs recently too, because on Ledger Live we use a custom logger, which is not console.*

packages/hw-transport-node-hid/src/listenDevices.js
```diff
@@ -27,37 +37,93 @@ export default (
...
     }
-    setTimeout(checkDevices, delay);
   };
 
-  timeoutDetection = setTimeout(checkDevices, delay);
+  debug("Starting to monitor USB for ledger devices");
+  usbDetect.startMonitoring();
+
+  // Detect add
+  usbDetect.on(`add:${VENDOR_ID}`, device => {
```
gre: https://github.com/LedgerHQ/ledgerjs/blob/master/packages/hw-transport-node-hid/src/isLedgerDevice.js

maybe we should filter like it was here?

packages/hw-transport-node-hid/src/listenDevices.js
```diff
@@ -27,37 +37,93 @@ export default (
...
+    debug("Device add detected:", device.deviceName);
```
gre: and this device don't have the path ? 😢

packages/hw-transport-node-hid/src/listenDevices.js
```diff
@@ -27,37 +37,93 @@ export default (
...
+          }, newDelay);
+        } else {
+          debug(`Giving up after ${attempt} attempts`);
+        }
       }
+    } else {
+      debug(`Polling skipped, retrying in ${delay}ms`);
 
-      lastDevices = currentDevices;
+      setTimeout(() => {
```
gre: this timeout does not get cleared if the whole listen is stop()-ed

packages/hw-transport-node-hid/src/listenDevices.js
```diff
@@ -27,37 +37,93 @@ export default (
...
+        poll(type);
```
gre: we have to poll multiple times because event happen too soon? Is it always the case, if so it would be great to estimate a minimal time and do an initial timeout to likely never have to poll twice. 🤔  

I think it would maybe preferable to refactor things using lodash `debounce` . each time there are remove/add events you just "push to 50ms later" the re-poll() , and the else case also is about triggering a poll().

e.g.

`poll = debounce(() => { }, delay)`

not sure what delay will be , if it make sense to be the same 1s that was before 🤔

packages/hw-transport-node-hid/src/listenDevices.js
MortalKastor: This is a pre-filter, taking advantage of the vendor ID filtering offered by `usb-detection`, our `isLedgerDevice` is used by `getDevices` afterwards

packages/hw-transport-node-hid/src/listenDevices.js
MortalKastor: No, it doesn't 😭 MadLittleMods/node-usb-detection#19

packages/hw-transport-node-hid/src/listenDevices.js
MortalKastor: > we have to poll multiple times because event happen too soon?

From my tests I have to poll a second time (after 500ms) only for device removal on macOS and Windows, but I have no idea how much this behavior is consistent from machine to machine 🤔

packages/hw-transport-node-hid/package.json
```diff
@@ -26,7 +26,9 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@ledgerhq/hw-transport": "^4.19.0",
-    "node-hid": "^0.7.2"
+    "lodash.debounce": "^4.0.8",
```
gre: it's a lil detail, but I would prefer we depend on `lodash` and doing `import debounce from "lodash/debounce"` , like on live

packages/hw-transport-node-hid/src/listenDevices.js
gre: 👌

packages/hw-transport-node-hid/src/listenDevices.js
gre: 🆗

gre:
approach i suggest to take: 
- let's do a beta release from this branch (before merge) – see the main README on how to do 
- we stress test it directly on ledger live (on 3 OS s, multiple devices, using a non dev build too)
- if all is fine, we can merge

gre:
if usb appears to be a pain to setup for windows, i suggest we go back to usb-detect, but manually manage the global event listening. for now, we could just listen always, because in LL we do this anwyay. just need to emit the events in some event emitter maybe, so each listen() can independently be created and unsubscribed.

packages/hw-transport-node-hid/src/isLedgerDevice.js
```diff
@@ -1,6 +0,0 @@
-// We check the usagePage on Win/OSX, and interface on Linux for get only HID device
-export default d =>
-  (["win32", "darwin"].includes(process.platform)
-    ? d.usagePage === 0xffa0
-    : d.interface === 0) &&
```
gre: ^ this wasn't a problem on linux? to filter the potential multi devices?

MortalKastor: My bad.
I restored the filtering.

gre:
let's all test together one more time on Linux, Windows and Mac **and** in Production build mode

gre:
let's rename `setListenDevicesPollingInterval` to `setListenDevicesDebounce`

packages/hw-transport-node-hid/src/isLedgerDevice.js
MortalKastor: `0xF1D0` is specific to FIDO (👀) U2F which is, indeed, purposefully filtered out so we don't list devices with the `U2F` app open, or a currency app with `Browser Support` set to on.

filoozom: Oh, so I need to disable browser support when using `hw-transport-node-hid`, I see! Sorry, the documentation is a bit scarce and I'm a newbie 😉. Thanks!

PR #192, opened 2018-07-24
Fix Zcash serialization

btchip:


gre:
it seems to affect a few "recorded" tests like `testBtcSegMulti`,
basically meaning the "snapshot" of the APDU have changed,
 i've not tested myself if test work on device.

is that supposed to change these apdu and if yes, any idea if this would help fixing bugs we have beside ZCash?

NB: contextual issue https://github.com/LedgerHQ/ledger-live-desktop/issues/1277

gre:
regression on Litecoin => `Missing inputs`

PR #194, opened 2018-07-25
Make `yarn build | clean | watch` work on Windows for each package

MortalKastor:
And keep current shell's `PATH` clean


meriadec:
good to know :p

PR #212, opened 2018-08-31
Tezos (Review Only)

jhartzell42:


gre:
@matthewbauer we'll hold this PR a bit until we converge to the same api:

fyi: we have a fork of the JS lib currently living in https://github.com/LedgerHQ/ledger-live-common/blob/getTransactionStatus/src/families/tezos/hw-app-xtz/index.js 

- code was rewritten a bit in a more concise / es style. 
- it also extract the `chainCode` info that Ledger Live (lib-ledger-core dep under the hood) need at the moment. unfortunately that means we need a forked version of the nano app too at the moment which is what we are developing with until nano app team and lib core team agree on the way to go.

qiluge:
hello, where can I find the document of ledger-xtz? Could you please give me some example code or document like https://github.com/LedgerHQ/ledgerjs/tree/master/packages/hw-app-btc? Thanks!

gre:
We've released https://github.com/LedgerHQ/ledgerjs/blob/master/packages/hw-app-tezos/src/Tezos.js back into ledgerjs. If there are work needed from this PR to be put back into this hw-app-tezos, feel free to send another PR. I'm sorry that this PR took that long to review.

PR #220, opened 2018-10-03
Support for unique scrambleKeys in coin constructors

schnogz:
### Description
Adds the ability to pass an optional scrambleKey to each coin type constructor classes.  If no scrambleKey is provided, it will fallback to the default key.

### Usage
```
const btc = new AppBtc(transport, 'K3Y');
```


gre:
(ignore the CI, seems something broke)

packages/hw-app-ada/src/Ada.js
```diff
@@ -53,15 +53,15 @@ export default class Ada {
   transport: Transport<*>;
   methods: Array<string>;
 
-  constructor(transport: Transport<*>) {
+  constructor(transport: Transport<*>, scrambleKey: ?string) {
```
gre: I suggest to move to default param syntax:

```
scrambleKey: string = "ADA"
```

that way you can directly pass `scrambleKey` down (and it's not an option since it get defaulted)

schnogz: good call 👍

PR #222, opened 2018-10-08
Make transport hid actually logging the APDU instead of read/write

gre:
also add a way to hook to a global debug function, for debugging a production app purpose.

Fixes #222

### DIFF
--- a/packages/hw-transport-node-hid/src/TransportNodeHid.js
+++ b/packages/hw-transport-node-hid/src/TransportNodeHid.js
@@ -229,7 +229,15 @@ export default class TransportNodeHid extends Transport<string> {
       return response;
     }
 
+    const { debug } = this;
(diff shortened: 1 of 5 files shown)


packages/hw-transport-node-hid/src/TransportNodeHid.js
```diff
@@ -229,7 +229,15 @@ export default class TransportNodeHid extends Transport<string> {
       return response;
     }
 
+    const { debug } = this;
     const deferred = defer();
+    if (debug) {
+      debug("=>" + apdu.toString("hex"));
```
meriadec: aaaaand I missed that missing space :scream:

gre: oops

PR #228, opened 2018-10-18
fix(Babel): rm transform-runtime

plondon:


gre:
@plondon i'll merge this and release a new beta for all libs to make sure there is no consequence for them.

rsmylskibc:
The removal of `babel-runtime` makes the packages unusable with any react app based off `create-react-app` `v 1.x`.

Apps now throw `ReferenceError: regeneratorRuntime is not defined`

Could this be reverted, or possibly have it changed to a more targeted babel config if there is one?

gre:
even if you do `import "babel-polyfill"` at first entry point of your app?

rsmylskibc:
Thanks - I was sure we had tried that and hadn't resolved it, but it is working now.

gre:
yeah it's a annoying restriction. I'm still not sure what's the correct answer. we'll also try to use babel 7 in a near future

menduz:
Any plan in reverting this?

gre:
Current trade-offs is to do this in your side: `import "babel-polyfill"`

menduz:
That's not sound :/

gre:
`transform-runtime` is creating problem with babel 7 so it's not a good solution neither.

this trade-off allow to make the library works both on babel 6 and babel 7 from my understanding. I'm happy to find a better solution, not sure if there are any consensus on this in JS libraries

menduz:
A possible fix could be to increase the compatibility level so `babel-polyfill` is no longer required, I believe that is possible tweaking the presets -speaking from ignorance with babel-. The browsers are good enough nowadays to not use major code transformations. I think if the browser has the APIs you require to talk to the devices, they do support generators and classes OOTB.

I beliebe typescript is better for libraries, because it doesnt require babel or any external polyfill and you can target the version of js you want without cross dependencies, you can even use async await in ES3. But I know that that requires a huge migration effort that may be impossible. Please take it into consideration for future libraries.

gre:
definitely would consider TS for 2019

webmaster128:
I think you can use babel-runtime if it adds value. You just need to make it a runtime dependency for every package you publish, not an (unpublished) dependency in the lerna root.

Issue #232, opened 2018-10-22
"Fees: UNKNOWN" displayed when signing non-segwit tx with utxos from segwit tx.


destrys:
### Gear
* Ledger Nano S - Firmware 1.4.2, MCU 1.5
* App: Bitcoin and Bitcoin Test v1.29
* Mac OS 10.13.5
* Chrome 69.0.3497.100
* @ledgerhq/hw-btc-app v4.24.0
* @ledgerhq/hw-transport-u2f v4.24.0

### Scenario
* Signing a BTC testnet transaction with `signP2SHTransaction`
* The single utxo being spent is a legacy P2SH multisig address
* The transaction funding the utxo is a segwit transaction
* Because the utxo being signed is non-segwit, the resulting signed transaction should be non-segwit
* During signing confirmation (on-device), the display first shows the Output #1 amount and destination address.
* After confirming the amount and destination, the ledger device displays "Fees: UNKNOWN" and requests a second button confirmation.
* An example script with an on-chain transaction and public WW is available at https://github.com/destrys/ledger-btc-signing-example/blob/master/scripts/sign_one_utxo.js
* I can confirm the same behavior with multiple utxo from segwit transactions.
* I have not tested behavior with non-segwit input transactions

### Solution
@btchip provided this fix via the ledger dev slack:

```trustedInputs.push({
            trustedInput: false,
            value: segwit
              ? Buffer.from(trustedInput, "hex")
              : Buffer.from(trustedInput, "hex").slice(4, 4 + 0x24),
            sequence
          });
```

> That part is ledgerjs might be there for compatibility with the older firmware versions - can you try to build a custom one removing the .slice and test again ?
> also change trustedInput : false to trustedInput : segwit ? false : true

I tested the following locally, and the ledger device correctly displayed the fees for the transaction.
```
trustedInputs.push({
            trustedInput: segwit ? false : true,
            value: segwit
              ? Buffer.from(trustedInput, "hex")
              : Buffer.from(trustedInput, "hex"),
            sequence
          });
```

monokh:
Also seeing this behaviour. From what the inputs are, there shouldn't be anything missing to calculate fees? In my case:
1 UTXO being spent
UTXO is legacy p2sh address (funding utxo is also same)
Is there a plan to address this @gre?

patwhite:
I'm seeing this with P2SH as well. Is there a way to just pass in what the fees are, we obviously have it calculated on our side.

[...]

PR #238, opened 2018-11-02
Fix BTC "Fees: UNKNOWN" bug

destrys:
This PR applies the fix described in #232 .

**Warning**: Tests are failing.  
I assume because this patch changes the data passed to the device (as expected).   
I'm not comfortable changing the test vectors, but am submitting the PR as recommended by @gre

### DIFF
--- a/packages/hw-app-btc/src/Btc.js
+++ b/packages/hw-app-btc/src/Btc.js
@@ -878,10 +878,10 @@ btc.signP2SHTransaction(
             0
           );
           trustedInputs.push({
-            trustedInput: false,
+            trustedInput: segwit ? false : true,
             value: segwit
               ? Buffer.from(trustedInput, "hex")
-              : Buffer.from(trustedInput, "hex").slice(4, 4 + 0x24),
+              : Buffer.from(trustedInput, "hex"),
             sequence
           });
         })


gre:
@bodymindarts could you check if this is a problem for you to change this part as you were the author of that specific code? https://github.com/LedgerHQ/ledgerjs/commit/ab1c09b5313b7b0f476686b0179002fe5b8ab7ae thanks

bodymindarts:
I can certainly look into this change. Can I have more context as to what is being achieved here? Why is this change necessary?

bodymindarts:
As far as I can tell this code change doesn't effect the code path that I added (where `segwit = true`).

Rather this is altering the behaviour of the default code path that was present since the file got introduced in [this commit](https://github.com/LedgerHQ/ledgerjs/commit/26c580cc#diff-5744b09e16f02f6ce77f7af7b90f628fR629).

destrys:
This change was recommended by @btchip 
The bug is that when I'm signing inputs from segwit transactions, but the address I'm spending from is legacy P2SH, the ledger displays "Fees: UNKNOWN". If I set segwit=true, the fee was displayed, but the signature is invalid. This change seemed to fix the problem, but I didn't rigorously test other scenarios.

packages/hw-app-btc/src/Btc.js
```diff
@@ -955,10 +955,10 @@ btc.signP2SHTransaction(
             0
           );
           trustedInputs.push({
-            trustedInput: false,
+            trustedInput: segwit ? false : true,
```
gre:
```suggestion
            trustedInput: !segwit,
```

gre:
it seems it changes the existing tests payload. i'll re-gen the tests if we confirm everything looks good on our end

humanumbrella:
Any update on this?

Matt561:
This bug is still present in current release when signing with Legacy P2SH.

landabaso:
The fix proposed by @destrys still works. I just tested it.
It is pretty straight forward.
This problem is here since long time ago.
Why was the problem for not merging it @gre ? Is it unsafe to apply it for some reason?

PR #239, opened 2018-11-08
Decred Integration

KhalilBellakrid:


packages/hw-app-btc/src/Btc.js
```diff
@@ -136,7 +136,8 @@ export default class Btc {
 
   getTrustedInput(
     indexLookup: number,
-    transaction: Transaction
+    transaction: Transaction,
+    isDecred?: boolean = false
```
gre: Prefer to pass additionals here and look for decred string because we've seen this doesn't scale

KhalilBellakrid: I agree ! I made requires changes

gre:
you might want to run `yarn prettier`

gre:
ok nevermind, test pass

packages/hw-app-btc/src/Btc.js
```diff
@@ -288,7 +310,7 @@ export default class Btc {
     inputs: Array<{ trustedInput: boolean, value: Buffer }>,
     bip143?: boolean = false,
     overwinter?: boolean = false,
-    additionals: string[] = []
+    additionals?: Array<string>
```
gre: i didn't see this in previous review, but why not keeping like before?

`additionals: string[] = []` guarantee the array will be defaulted to `[]` when not provided. any reason to switch to `?: ` again?

PR #247, opened 2018-11-20
Cleanup babel config

### LINKED ISSUE #211: Missing babel-runtime dependency in packages
It seems that `@ledgerhq/hw-transport-node-hid` is requiring `babel-runtime`, but it is not included as a dependency in that particular `package.json`.

### Steps to reproduce
- Running the _Basic example_ without `babel-runtime`, as for example in [ledger-btc-test](https://github.com/Wollac/ledger-btc-test).
- Manually adding the dependency `babel-runtime@6.26.0` to the executing project resolves the issue.

### Error
```
Error: Cannot find module 'babel-runtime/core-js/get-iterator'
```

webmaster128:
* After the [transform-runtime plugin was removed](https://github.com/LedgerHQ/ledgerjs/pull/228), we don't need the dependency anymore.
* There are no runtime dependencies for the lerna monorepo root; Move them to devDependencies to make this clear

I verified that there is no `require("babel-runtime)` in the libs/ output anymore, ~thus closes #211~ but all packages need a rebuild

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,12 +25,10 @@
   "devDependencies": {
     "babel-cli": "^6.26.0",
     "babel-eslint": "^8.0.2",
-    "babel-plugin-transform-runtime": "^6.23.0",
     "babel-preset-env": "^1.7.0",
     "babel-preset-flow": "^6.23.0",
     "babel-preset-react": "^6.24.1",
     "babel-preset-stage-0": "^6.24.1",
-    "babel-runtime": "^6.26.0",
     "browserify": "^16.2.3",
     "create-hash": "^1.1.3",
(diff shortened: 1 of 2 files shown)


webmaster128:
Is there anything blocking this from being merged? Right now, babel-runtime is obsolete. If you decide to revert #228, you need

* Revert #228
* Add "babel-plugin-transform-runtime" to lerna root devDependencies
* Add "babel-runtime" as a runtime dependency to every published package

In any case, this cleanup makes sense I think.

gre:
nothing is blocking but time for the team to test this. i'll take a look as soon as I have more time 🤞

PR #253, opened 2018-11-30
add RSK derivation options

scmilee:
In a similar spirit to https://github.com/trufflesuite/truffle-hdwallet-provider/pull/64. I was hoping to add the ability to derive RSK specific addresses from this sub-provider. RSK's [utility](https://utils.rsk.co/) mimics the derive path "44'/0'" with a process that matches[Ethereumjs-tx](https://github.com/ethereumjs/ethereumjs-tx) and  [Bip-39](https://iancoleman.io/bip39/) derivations when converting BTC private keys into RSK private keys as shown [here](https://utils.rsk.co/).


scmilee:
Tests are failing, likely with parameter guards against different derivation paths, will work on fixing them.

PR #262, opened 2018-12-19
Allow unsignable inputs to be included in createPaymentTransactionNew

plondon:
Due to some network forks not providing replay protection it is vital to allow users to split their coins so they can be used on either chain. If user includes coin dust from one chain in part of their transaction, it cannot be used on the other forked chain.

This PR allows the client to pass an additional array of inputs that can be signed by another service, but not the device.

### DIFF
--- a/packages/hw-app-btc/src/Btc.js
+++ b/packages/hw-app-btc/src/Btc.js
@@ -715,15 +715,15 @@ btc.createPaymentTransactionNew(
                 value: Buffer.from(trustedInput, "hex"),
                 sequence
               });
-              let offset = useBip143 ? 0 : 4
+              let offset = useBip143 ? 0 : 4;
               targetTransaction.inputs.push({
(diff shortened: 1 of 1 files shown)


packages/hw-app-btc/src/Btc.js
```diff
@@ -697,6 +699,32 @@ btc.createPaymentTransactionNew(
           });
         }
       })
+      .then(() => {
+        return foreach(multisigInputs, input => {
+          return doIf(!resuming, () =>
+            getTrustedInputCall(input[1], input[0]).then(trustedInput => {
```
amougel: Don't we want to pass the additionals here?

plondon: Good point, btw if you want to test this actually works you could try running this branch of the wallet locally https://github.com/blockchain/blockchain-wallet-v4-frontend/pull/1306. I can send you some bch as well.

PR #267, opened 2019-01-11
Fix transaction signing with lockTime and expiryHeight (ZEC, KMD sapling)

smdmitry:
Current BTC app version handles transactions with non zero locktime and expiryHeight) incorrectly.
Signs invalid inputs hex and returns invalid transaction hex.

I fixed it for ZEC and KMD coins. Didn't check other coins, from the code I can see that Decred can be affected.


gre:
thanks for the PR!

PR #271, opened 2019-01-24
Revert PR of multisigInputs feature

KhalilBellakrid:
This PR reverts commits a843a7a9f4 ... ac0e3bb15c related to multisigInputs feature which was causing errors in Live when signing bitcoin like transactions (error `Invalid Data 0x6f...`)
It includes also fix of Decred that was failing at broadcast because of oprhan transactions, bug was occurring when creating getting trusted inputs: streamed data was wrong, `tree` field was incorrect

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,7 +3,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "4.33.6",
+  "version": "4.33.7",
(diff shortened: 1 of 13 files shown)


KhalilBellakrid:
This is replaced by : https://github.com/LedgerHQ/ledgerjs/pull/272

PR #285, opened 2019-02-04
LL-898 Added error for cryptos not supporting source===destination

juan-cortes:
Needed for LL-898 for currencies not supporting sending funds to the origin account such as xrp


juan-cortes:
@gre can't request review from you nor merge

PR #284, opened 2019-02-02
Introduce devices and refactoring of transports

gre:
- introduces @ledgerhq/devices package to deal with different devices and share some information.
- #281 close() now waits that the exchange() has finished in the returned promise
- #280 there is a `deviceModel` field in the event of `listen()`
- React Native Bluetooth: fixes "Operation was cancelled" bug in Samsung S9
- #279 web-ble implementation implements a workaround for a bug in current NanoX firmware
- full refactoring of node-hid implementation from the work done in webusb implementation. Much simpler code and more robust. Also using the HID channels.
- factorization of the "atomic" logic used in many hw-transport implementations.
- web-ble implementation loads the MTU at open() to benefit of better performance.
- #69 node-hid transport implementation emits a "disconnect" event when a write/read throw a "cannot write" or "cannot read" error.
- `packages/documentation-website/` is removed because not maintained. we'll rewrite proper documentation soon.

... more to detail...

...to be continued...

## Deprecated

- `TransportError` and `TransportStatusError` should be imported from `@ledgerhq/errors`

## Breaking changes

none

## Repository changes

- complete rewrite of the README.md
- examples are moving to https://github.com/LedgerHQ/ledgerjs-examples
- documentation project is removed, we'll write better documentation in each package README

---

## Remaining to do

- heavily test ledger-live-desktop with it.
- heavily test ledger-live-mobile with it.

## Later

- share more code of BLE framing implementation
- TransportError: might replace it entirely with better precision error. maybe only keep it for protocol errors.
- @ledgerhq/react-native-hid => @ledgerhq/hw-transport-react-native-hid ?
- @ledgerhq/react-native-hw-transport-ble => @ledgerhq/hw-transport-react-native-ble ?

PR #286, opened 2019-02-06
Implement 'Provide ERC 20 token information' incoming feature

gre:
- A script that loads the ERC20 data from our internal crypto-assets db.
- the compressed data of all supported ERC20 committed in the lib `@ledgerhq/hw-app-eth`
- `@ledgerhq/hw-app-eth/erc20` optional dependency that exposes a `byContractAddress` and `list` to access the data
- A new method in the `@ledgerhq/hw-app-eth` app: `provideERC20TokenInformation` that can just take the result of `byContractAddress(addr)`

see doc https://github.com/LedgerHQ/ledger-app-eth/blob/externalize-erc20/doc/ethapp.asc

## Example

```js
import AppEth from "@ledgerhq/hw-app-eth"
import { byContractAddress } from "@ledgerhq/hw-app-eth/erc20"
const appEth = new AppEth(transport)
const zrxInfo = byContractAddress("0xe41d2489571d322189246dafa5ebde1f4699f498")
if (zrxInfo) await appEth.provideERC20TokenInformation(zrxInfo)
const signed = await appEth.signTransaction(path, rawTxHex)
```

```
e00a000066035a5258e41d2489571d322189246dafa5ebde1f4699f4980000001200000001304402200ae8634c22762a8ba41d2acb1e068dcce947337c6dd984f13b820d396176952302203306a49d8a6c35b11a61088e1570b3928ca3a0db6bd36f577b5ef87628561ff7
```

## Notes

the currently data size of erc20 is 155 kb of uncompressed JavaScript. This will increase a bit your bundle size but **only** if you effectively use `"@ledgerhq/hw-app-eth/erc20"`.
This allows to be run offline but it's up to you to chose to not use it client side and host/serve it yourself.


gre:
(ignore circle CI that is only here for the other PR #284)

packages/hw-app-eth/src/erc20.js
```diff
@@ -0,0 +1,80 @@
...
+  data: Buffer
+};
+
+export type API = {
+  byContract: string => ?TokenInfo,
+  list: () => TokenInfo[]
+};
+
+const asContractAddress = (addr: string) => {
+  const a = addr.toLowerCase();
```
meriadec: this case thing has been an issue and discussed on Vault: should the front must handle this conversion, assuming the contract address has wrong case, or do we trust our unique source (which is crypto-assets-list)

we took the decision to never touch the contract address.
maybe it's not the exact same problem here, but I just put a warning on opening the door of manipulating the address.

gre: we need to keep it here because this lib is to give to the world.
e.g. MEW want to integrate the support of that, maybe they have a uppercase addr? we don't know, we need to converge to the same format. same for the 0x vs not having 0x.

PR #288, opened 2019-02-06
LL-980 Moved the exchange logic to a separate thread

juan-cortes:
Allows exchanges with the devices connected via usb otg without blocking the app's thread.


juan-cortes:
@KhalilBellakrid @pollastri-pierre  can you please take a look at this, we are concerned that there might be an associated cost at taking this approach since some operations create many apdu exchanges and this would be creating threads for each of those exchanges. My understanding is that after the run method is completed the thread is finished, there's no need to gc but is the overhead cost of doing this negligible/should we worry?

I played a bit with a version using a thread pool of 1 to avoid creating unnecessary threads and I haven't noticed any speed difference at all (I also had some issues with connecting back to the dongle in our modals after a dismiss, but maybe that can be resolved after https://github.com/LedgerHQ/ledger-live-mobile/pull/745 is merged)

gre:
I was suggesting a thread pool of size 1. But need an expert 👀

pollastri-pierre:
You can use a single queue executor for this https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executors.html#newSingleThreadExecutor()

juan-cortes:
Moving to `SingleThreadExecutor` and having the permission pr merged made it work as expected. This means we are no longer blocking the app, I don't know how easy it would be for @Arnaud97234 to test this though, wdyt @gre

gre:
@juan-cortes 
- release a beta version from that PR directly (documented here: at the end, "canary" https://github.com/LedgerHQ/ledgerjs#deploy )
- bump on the mobile side

PR #289, opened 2019-02-07
Upcoming Cardano integration

ppershing:
This PR updates Cardano implementation to be in sync with upcoming integration of https://github.com/vacuumlabs/ledger-cardano-app

Ledger questions
- we depend on `base-x` package, is it ok?
- `yarn.lock` - I am not sure why prettier was added. On the other hand, I wonder why it wasn't locked before our PR
- are we allowed to do the `.babelrc` change?
- we created `example-node-ada`, I hope that is fine

VacuumLabs check points (should be checked before merging)
- [ ] ensure we have good doc comments
- [ ] check yarn lint / flow


ppershing:
Closing this PR as it was agreed to move the package into its own repo (https://github.com/cardano-foundation/ledgerjs-hw-app-cardano)

PR #299, opened 2019-02-21
Added support for ledger live derivations

### LINKED ISSUE #200: Add more supported derivations for web3-subprovider
For now @ledgerjs/web3-subprovider support only `44'/60'/x'/n` pattern.
**But in Ledger Live they're using `m/44'/60'/0'/0`, so please support this derivation in @ledgerjs/web3-subprovider.**

p/s: Could @ledgerjs/web3-subprovider accept a lot of derivations like myetherwallet.com doing?

### LINKED ISSUE #273: Inconsistent derivation  between web3-subprovider and Ledger Live application
Ledger Live application uses 44'/60'/0'/0/0 derivation, but web3-subprovider allows only to use something like 44'/60'/0'/0, which as the result generates different address.

peteremiljensen:
Currently, Ledger Live uses different HD derivations paths, which results in inconsistent addresses. The issue is fixed by simply changing the path validation.

Closes #200 and closes #273 

Co-author @truls

### DIFF
--- a/packages/web3-subprovider/src/index.js
+++ b/packages/web3-subprovider/src/index.js
@@ -15,12 +15,12 @@ function makeError(msg, id) {
 }
 
 function obtainPathComponentsFromDerivationPath(derivationPath) {
-  // check if derivation path follows 44'/60'/x'/n pattern
-  const regExp = /^(44'\/(?:0|1|60|61)'\/\d+'?\/)(\d+)$/;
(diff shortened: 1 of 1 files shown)


gre:
thanks.

in the future, we'll probably also generalize this so there is no longer this path restriction.
for now let's take this :)

peteremiljensen:
Sounds good 👍

PR #303, opened 2019-02-26
LL-1086 Broadcast connect/disconnect events from react-native-hid

juan-cortes:
This listens to devices being plugged in/out of the phone and broadcasts the event. 

~~Since we don't have a unique identifier to selectively remove a device from the list I've added the `reset` event, which is not handled perfectly (it adds some stuff) from live-common but doesn't affect the functionality. Also, need to handle that `reset` on mobile end to delete the list of devices added since we retrigger the list.~~

We merge an initial list of connected devices with a subject that will emit subsequent connections and disconnects, this eliminates the need to unsubscribe and support multiple instances of the device selector running at the same time, with all receiving the broadcasted events.

### DIFF
--- a/packages/react-native-hid/android/src/main/java/com/ledgerwallet/hid/ReactHIDModule.java
+++ b/packages/react-native-hid/android/src/main/java/com/ledgerwallet/hid/ReactHIDModule.java
@@ -15,6 +15,7 @@ import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReactContextBaseJavaModule;
+import com.facebook.react.modules.core.DeviceEventManagerModule;
 import com.facebook.react.bridge.ReactMethod;
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.WritableArray;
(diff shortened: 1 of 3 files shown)


packages/react-native-hid/android/src/main/java/com/ledgerwallet/hid/ReactHIDModule.java
```diff
@@ -29,30 +30,67 @@
...
+
+                    UsbDevice device = (UsbDevice) intent.getExtras().get(UsbManager.EXTRA_DEVICE);
+                    getReactApplicationContext()
+                            .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class)
+                            .emit("onDeviceConnect", buildMapFromDevice(device));
+                        break;
+                    case ACTION_USB_DETACHED:
+                        getReactApplicationContext()
+                                .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class)
+                                .emit("onDeviceDisconnect", null);
```
gre: in this context we don't know *who* disconnects? we can't do the same buildMapFromDevice?

packages/react-native-hid/src/index.js
```diff
@@ -9,6 +9,34 @@ type DeviceObj = {
 };
 
 export default class HIDTransport extends Transport<DeviceObj> {
+  static observer: *;
+  static unsubscribed: boolean;
```
gre: I would prefer these are not static in the class but private global variables. So they are not exposed outside.

packages/react-native-hid/src/index.js
```diff
@@ -9,6 +9,34 @@ type DeviceObj = {
...
+      });
+    }
+  };
+
+  static onDeviceDisconnect = (_: *) => {
+    if (HIDTransport.observer && !HIDTransport.unsubscribed) {
+      HIDTransport.observer.next({ type: "reset" });
+      HIDTransport.listDevices();
+    }
+  };
```
gre: same for the 2 callbacks. can be functions outside the class.

packages/react-native-hid/src/index.js
```diff
@@ -18,20 +46,27 @@ export default class HIDTransport extends Transport<DeviceObj> {
     return list.filter(d => d.vendorId === ledgerUSBVendorId);
   }
 
-  static listen(observer: *) {
-    if (!NativeModules.HID) return { unsubscribe: () => {} };
-    let unsubscribed = false;
+  static listDevices(){
```
gre: inline this in the listen itself.

packages/react-native-hid/src/index.js
```diff
@@ -18,20 +46,27 @@ export default class HIDTransport extends Transport<DeviceObj> {
...
           const deviceModel = identifyUSBProductId(c.productId);
-          observer.next({ type: "add", descriptor: c, deviceModel });
+          HIDTransport.observer && HIDTransport.observer.next({ type: "add", descriptor: c, deviceModel });
         }
       }
     });
+  }
+
+  static listen(observer: *) {
+    HIDTransport.observer = observer;
```
gre: Please add a failsafe that throw an error if `HIDTransport.observer` is still not `null` . otherwise we will have ugly race condition since the code provide here don't allow to run two listen at same time.

warning: we maybe have a problem in the current mobile app if you open the Receive flow when also being in Manager? 🤔

**Here is what I suggest:**
- after you add this failsafe (once at a time), If there is no problem with current mobile app. fine, let's ship this.
- otherwise, we should support multiple listen(). To do that, I recommend the use of rxjs. Using some Rx.Subject to broadcast the events.

packages/react-native-hid/android/src/main/java/com/ledgerwallet/hid/ReactHIDModule.java
juan-cortes: Yes. but the id ends up being the same for two different devices on the same port. I plug my personal nano s and one of my dev nano s, it will not know which one is which. All this is because of a theoretical support for multiple dongles which will likely never happen

packages/react-native-hid/src/index.js
juan-cortes: ok

packages/react-native-hid/src/index.js
juan-cortes: ok

packages/react-native-hid/src/index.js
juan-cortes: ok

packages/react-native-hid/android/src/main/java/com/ledgerwallet/hid/ReactHIDModule.java
juan-cortes: For 1 device support this works..

juan-cortes:
@gre ready for a second review

Issue #307, opened 2019-02-27
Ethereum derivation path for Ledger Live calculated incorrectly


vevnin:
In case when user creates multiple Ethereum accounts in Ledger Live 1.4.1 derivation path for each of them is set in pattern described in [BIP 44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) incrementing account field:
- 44'/60'/0'/0/0
- 44'/60'/1'/0/0
- 44'/60'/2'/0/0

Web3-subprovider now supports 44'/60'/0'/0/0 path, but when it performs scan for addresses it generates paths incrementing address_index field instead:
- 44'/60'/0'/0/0
- 44'/60'/0'/0/1
- 44'/60'/0'/0/2

This way user of dapp where Web3-subprovider is used can see only first address from Ledger Live and can't see the rest of claimed addresses.

vevnin:
#273

gre:
Yes as you said, the BIP44 is favorized in current implementation of Ledger Live derivation. This is not likely going to change.

However, Ledger Live **does** scan for `44'/60'/0'/0/{x}` accounts, so these should be discovered! meaning you should be able to move eths to these accounts and it will work in Ledger Live.

Maybe in the future, we can offer in Ledger Live a preference to chose the derivation. If someone want to offer a new Settings in Ethereum currency settings, be my guest!

vevnin:
@gre The issue is not that Ledger Live doesn't detect 44'/60'/0'/0/{x} accounts. It's rather related to web3-subprovider and how it performs account scan.

In case if account is created first in Ledger Live, it gets created with 4'/60'/{x}'/0/0 path. web3-subprovider scans using  44'/60'/0'/0/{x} pattern and it **does not** detect any accounts created in Ledger Live besides first one. It basically gives you different set of addresses starting from second one.

This way users see mismatch between accounts displayed in Ledger Live and in dapp using web3-subprovider.

gre:
OH yeah sorry! I thought you were mentionning an external library!

you are correct.

gre:
this is now fixed (version `5.2.0`)

you can do this:

```
paths: ["44'/60'/x'/0/0"]
```

or


```
paths: ["44'/60'/0'/x"]
```

and even


```
paths: ["44'/60'/x'/0/0", "44'/60'/0'/x"]
```

[...]

PR #312, opened 2019-03-04
 Ethereum derivation path is calculated same way as in Ledger Live

vevnin:
This PR fixes issue described in #307 "Ethereum derivation path for Ledger Live calculated incorrectly"


gre:
I think the code is getting a bit too hairy at this point, ideally we probably should just provide the scheme to follow.

like:

`44'/0'/{x}'/0/0` and then we can just `.replace('{x}', x)` during the iteration. It would simplify a lot the logic. To me we can also remove all the code that was checking for "allowedHdPaths". We just allow everything, worse case scenario is that the device will error in case path ain't allowed.

this is going to be a breaking change if overriding the `path` argument, but we can just `throw` an explicit error if there is no `{x}` in that path.


@vevnin I'll close this PR in favor of having this simpler idea implemented.
feel free to implement this if you want, otherwise i'll refactor this in a few days/weeks.

I hope this make sense. Thanks

PR #340, opened 2019-04-12
fix: LL-1174 Better error on device disconnect retry on AccountAdd (anywhere really)

juan-cortes:
When retrying a device interaction that failed the error was no longer referring to the disconnection but rather to the null reference to the device. This maps that error and shows the device disconnect error. **Only affects wired dongles to mobile.**

There are several more edge cases like this one that will eventually need to be addressed


gre:
@juan-cortes can you `prettier` that `}catch(error){` ? :p

juan-cortes:
Done

PR #344, opened 2019-04-17
Rework hw-transport-mocker and tests

gre:
## new feature in command `ledger-hw-http-proxy-devserver`

record apdus into a file

```
RECORD_APDU_TO_FILE=1 ledger-hw-http-proxy-devserver -f apdus.log
```

replay apdus from this file

```
ledger-hw-http-proxy-devserver -f apdus.log
```

## improve the test

each snapshot is saved into snapshots/* file

it's just a series of apdu (with `=> ` and `<= `) that get replayed in a queue.


packages/hw-transport-mocker/src/RecordStore.js
```diff
@@ -1,41 +1,108 @@
...
-    const { cache } = this;
+    this.queue.push([apdu.toString("hex"), out.toString("hex")]);
+  }
+
+  replayExchange(apdu: Buffer): Buffer {
+    const { queue } = this;
+    if (queue.length === 0) {
+      throw new RecordStoreQueueEmpty();
+    }
+    const [head, ...rest] = queue;
```
flocks: what about renaming `...rest` to `...tail`? :p

gre: done!

PR #347, opened 2019-04-25
Remove useless minSdk in AndroidManifest.xml

gre:
Remove useless (and now breaking) minSdk in AndroidManifest.xml


gre:
For now I'm sure it doesn't with it 🤪 in latest react native

Issue #352, opened 2019-05-16
NotSupportedError: GATT operation failed for unknown reason


harshjv:
I am using Ledger Nano X with;

|Lib/App|Version|
|-|-|
|Secure Element|1.2.4.1|
|Micro controller|2.8|
|Bitcoin App|1.3.8|
|Ethereum App|1.2.3|
|@ledgerjs libs|4.56.0|

Unfortunately, web-ble doesn't work in Chrome for Android v74. It throws this error;

```
NotSupportedError: GATT operation failed for unknown reason
```

Although same works with Chrome on Desktop.

Demo: https://cross-chain-playground.harshjv.com/wallet/btc/bitcoin (select Ledger Nano X on Chrome for Android)

gre:
Hi @harshjv, really cool demo BTW!

I'll give a try ASAP. Maybe something the current implementation doesn't do correctly for Chrome Android to support it 🤔

Were you able to make WebUSB working on Chrome Android? I think it works last time I tried.

jafri:
This issue is not just on Chrome Android, seeing same issue on all Android Browsers

jafri:
Any updates on this?

gre:
There is no update yet, the conclusion is this bug happens when you fall into a known bug with the current firmware, you need in that case to Reset Pairings on the NanoX and forget on the phone/computer side. You also should not install too much apps on the device (no more than 50 for instance) because we've seen it's more likely you fall into the problem.

This will require a new firmware.

ben-oxley:
Sorry to drop in on this un-invited. I came across this whilst encountering similar issues with web-bluetooth for a website I'm implementing. It worked well in Win 10 chrome, but had issues on Android Chrome throwing NotSupportedError. Anyway, the solution for me was to sort out my charateristic notification subscriptions to correctly return promises all the way through the chain and to make sure subscriptions happened serially rather than in parallel. If you want to see what I'm talking about, have a look here: https://github.com/SGSVertigo/vertigo-web/commits/master at commits 86fd91df96adbf52a6e2a56c2176103560cee5f9 and f62bad57d21a8ea0515ea7ceedf8d78018d175c4. My theory is that android fails when it is trying to execute notification subscriptions in parallel. I can't see where in your code base you're implementing it, but the symptoms sounds similar!

gre:
[...]

PR #353, opened 2019-05-16
Attempt to fix WebBLE on Android

gre:
might help on #352 bug


gre:
Sadly doesn't fix the bug. However unifying to the same timeout at the reactnative implementation can be good so ze can merge

PR #358, opened 2019-05-19
Use @ledgerhq/logs in ledgerjs libraries + commit flow-typed instead of runtime resolution

gre:
logs are centralized in library `@ledgerhq/logs` to make it easier to use across Ledger projects.


gre:
i'll rework the PR to split things out. won't do that flow-typed commit for now

PR #363, opened 2019-05-22
Use @ledgerhq/logs

gre:
logs are centralized in library `@ledgerhq/logs` to make it easier to use across Ledger projects.

- @ledgerhq/hw-transport (all transports): `setDebugMode` usage is deprecated and is now a noop
- @ledgerhq/hw-transport-node-hid: `setListenDevicesDebug` usage is deprecated and is now a noop
- imports on `"@ledgerhq/hw-transport-web-ble/lib/debug"` will need to be replaced by `@ledgerhq/logs` (same for the react-native version)

The way to get logs is by using `@ledgerhq/logs` library (doc: https://github.com/LedgerHQ/ledgerjs/tree/master/packages/logs )

APDU are logged as type `"apdu"`. This PR also add more logs like for instance the ability to see the hid frames as well.


nshCore:
Have you got any docs on how to use this, I appreciate the deprecation warning but there is no info on how to use this.

PR #375, opened 2019-07-03
hw-app-btc: Add signUtxo method

JonSalazar:
XSN allows coin staking by creating a TPoS contract. This contract is included
into a transaction as an extra OP_RETURN output, it requires a signature from a
collateral UTXO (or OutPoint), the signUtxo function generates that signature.

this change depends on https://github.com/LedgerHQ/ledger-app-btc/pull/120


FabriceDautriat:
The Bitcoin app modification isn't scheduled yet on our end.

FabriceDautriat:
We won't be merging this change for the time being.

PR #390, opened 2019-07-31
Document `createPaymentTransactionNew` "bech32" additionals param

monokh:
To spend native segwit outputs, `bech32` needs to be included in the additionals param. This wasn't documented.


gre:
I merged this too quickly: the doc is generated from the sourcecode js doc itself. we should later fix back this in the JS itself.

PR #393, opened 2019-08-19
LL-1711 Change default timeouts for transports

juan-cortes:
I can't really test this due to the nested dependencies between live-common, mobile, and the individual transports with hw-transport package itself, Metro bundler goes nuts. But I assume that these are the timeouts we want to modify in order to avoide the scenario where the user selects a nano x and has not enough time to even turn it on before a timeout (3 seconds!).

Also 60 seconds for the other timeout because that's what I wrote on Jira for some reason 🤷‍♂ 

https://github.com/LedgerHQ/ledgerjs/blob/master/packages/hw-transport/src/Transport.js#L211


juan-cortes:
This doesn't necessarily address the problem, closing for now.

PR #405, opened 2019-10-03
Add GasLessThanEstimate error for warning user of low gas

juan-cortes:
This error could be used to warn the user that the input gaslimit is lower than our estimation and that the transaction might fail with outofgas. The below wording is just an example, we could use this warning to show a banner, or a second popu like we do when fees are too high for instance.

image


gre:
great!

PR #427, opened 2019-11-26
treat 61xx status response as success

flocks:
61xx status are actually valid response, where xx is the length of the
remaning data. It's a way to paginate the response when it's too long.

### DIFF
--- a/packages/hw-transport/src/Transport.js
+++ b/packages/hw-transport/src/Transport.js
@@ -194,7 +194,11 @@ TransportFoo.open(descriptor).then(transport => ...)
       ])
     );
     const sw = response.readUInt16BE(response.length - 2);
-    if (!statusList.some(s => s === sw)) {
+
+    // 61XX is actually a success response according to ISO7816 where XX is the length of the remaining response to retrieve
+    // basically it's a way to paginate the response
+    const is61xxStatus = sw >= 0x6100 && sw <= 0x61ff;
+    if (!is61xxStatus && !statusList.some(s => s === sw)) {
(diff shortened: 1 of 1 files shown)


flocks:
Hmm I'm thinking may be I should directly handle the pagination (and aggregating the responses) directly in ledgerjs. What do you think @meridadec @gre ?

gre:
why not.

I actually wonder if we shouldn't take the current opportunity (migration to babel, major bump incoming) to do changes on that send and/or rethinking it a bit.
I wonder if the send shouldn't be done somewhere else, in some sort of utilities.

I think i don't like about this send is that it doesn't split the status from the data response.

What we could do is:

- not touching this current transport.send so we're actually not breaking changes. but we deprecate it and `console.warn` it clearly
- the `send` is moved into a function somewhere else. 

PoC:

```js
import { send } from "@ledgerhq/hw-transport"

const data = await send(transport, 0xE0, 0x01, 0, 0); // data would not contain the 9000 status part
const data = await send(transport, 0xAA, 0xBB, 0, 0, Buffer.from(...));
```

and to me with no longer the possibility to set the accepted status but just a built in impl of what you mentioned.
the fact it's a function would make it either for alternative impl instead of built in in hw-transport via weird inheritence (cc @IAmMorrow )

Fundamentally a transport is just about implementing the exchange function (+ some lifecycle, open,close). things around it are utilities that consume this.

flocks:
Ok so @meriadec I suggest we hardcore all the 61xx status on our side, so we pass them to the current `send` as `statusList` param so we have time to think about a better implementation of `send`

PR #435, opened 2020-01-02
hw-transport-node-hid-singleton: listen() unsubscribe to stop monitoring

### LINKED ISSUE #434: node-singleton implementation to release the usb listening
need to reboot this idea https://github.com/LedgerHQ/ledgerjs/blob/0453b2b6ed3702d81277a529b1651f8528d13878/packages/hw-transport-node-hid-singleton/src/listenDevices.js#L26-L41

because without doing stopMonitoring(), you will have a node thread never ending.

gre:
Fixes #434 
Blocked by https://github.com/MadLittleMods/node-usb-detection/issues/57

will need to be tested a bit

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,16 +39,16 @@
     "create-hash": "^1.1.3",
     "documentation": "^12.1.4",
     "eslint": "^6.8.0",
-    "eslint-config-prettier": "^6.0.0",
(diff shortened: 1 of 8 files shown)


gre:
does not seem to work properly. when we startMonitoring/stopMonitoring multiple times, i can see some seg faults...

Issue #451, opened 2020-01-25
web3-subprovider is broken after upgrading ethereumjs-tx to 2.1.2


sk91:
ethereumjs-tx does not expose default anymore, but `{Transaction}` instead

fixed in https://github.com/LedgerHQ/ledgerjs/pull/452

PR #452, opened 2020-01-25
fix(web3-subprovider): incorrect import of ethereumjs-tx

sk91:
fixes https://github.com/LedgerHQ/ledgerjs/issues/451


gre:
thanks a lot!

Issue #397, opened 2019-08-30
Publish ES6 Modules


MicahZoltu:
The libraries, including the web libraries, all appear to be compiled for CommonJS only.  Recommend including a `module` property in `package.json` for each of the packages that points at a build that targets ES6 modules.

gre:
definitely need to explore that in the next major rework of our build system (for the context we are also considering a major lib rework with typescript)

MicahZoltu:
I would love to see a move to TypeScript.  However, not being written in TS is a minor inconvenience that I can work around, while not having ES6 modules is notably more painful, so if I had my way I wouldn't delay ES6 support in lieu of TS support.

Mrtenz:
I would like an ES6 version of Ledger.js packages too. I'm using `@ledgerhq/transport-web-ble` for example, and currently it requires all of `rxjs` to be bundled. That's about 350 KB, over a third of my whole application. A rewrite to TypeScript is an added bonus.

gre:
@Mrtenz I'm not sure to understand why a ES6 module would help on the bundle size, the lib would still depends on rxjs.
I will take a look of it's possible to tweak the rxjs import to lower the bundle size but I'm not sure.

Mrtenz:
@gre I'm assuming Ledger.js is not using all features of Rxjs? Rxjs has an ES6 version, and if Ledger.js is an ES6 module too, Webpack should be able to do tree shaking and remove any unused parts of Rxjs. At least, I think. I'm using Babel to generate an ES module [here](https://github.com/FindETH/wallets/blob/ab91ec31ca25bbce0331f843cd48ab358af42d3b/package.json#L25), and in [my Webpack bundle](https://github.com/FindETH/web/tree/7f496262f738616fb2c11e2b6929bee1ab10ec87) any unused stuff is removed (you can try for yourself with `yarn build:analyze`).

350 KB seems rather large for an Observer library in the first place though. 😅

gre:
Thanks, that's useful i'll try on my end.

gre:
@Mrtenz could you tell me if `5.1.1-alpha.2` have fixed the problem? Using `parcel` I don't see any change but I suspect parcel does not use `"module"` entry point but always the `"main"`
thanks

Mrtenz:
@gre This reduced the size for Rxjs from about 320 KB to 156 KB! A 50% reduction just by using ES modules. I really appreciate you looking into this!

gre:
[...]

PR #467, opened 2020-03-08
Setup TypeScript with rollup.js

LukeSugiura:


LukeSugiura:
I'll create a different PR for dealing with loose types (i.e. "any)

packages/errors/package.json
```diff
@@ -16,18 +16,18 @@
   "publishConfig": {
     "access": "public"
   },
-  "main": "lib/index.js",
-  "module": "lib-es/index.js",
+  "main": "dist/index.js",
+  "module": "dist/index.js",
```
gre: these shouldn't be the same folder right? or maybe i miss something. 

my understanding is: essentially the "module" helps optimizing bundle size and should be the new way of doing things, main is here for backward compat.

LukeSugiura: I bundle the file only in umd format so I thought it works for both platform. But if you want, I can change it so that the bundler outputs several files with different format.

gre: preferably yes. this was a community request that we addressed recently https://github.com/LedgerHQ/ledgerjs/issues/397

LukeSugiura: I can't denied request from a community especially from those two 😄
fix: https://github.com/LedgerHQ/ledgerjs/pull/467/commits/04f9a934c368a8ce35e5635d20dfb7a4a8862873

package.json
```diff
@@ -7,33 +7,33 @@
   ],
   "scripts": {
     "preinstall": "node -e \"if (process.env.npm_execpath.indexOf('yarn') === -1) { console.log('\u001b[31mPlease use yarn\u001b[0m'); process.exit(1); }\"",
-    "test": "yarn run build && yarn run lint && yarn run prettier-diff && yarn run doc && jest",
+    "test": "yarn run build && yarn run lint && yarn run doc && jest",
     "test-no-git-diff": "git diff --exit-code || exit 1",
     "test-node": "cd packages/test && yarn run test-node",
     "test-browser": "cd packages/test && yarn run test-browser",
-    "prettier": "prettier --write 'packages/*/src/**/*.js'",
```
gre: this was accidentally removed. i'll re-add

PR #487, opened 2020-04-07
fix(hw-app-trx): use 250 bytes as chunk size

andelf:


andelf:
FYI @fbsobreira

packages/hw-app-trx/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ledgerhq/hw-app-trx",
-  "version": "5.12.0",
+  "version": "5.12.1",
```
gre, line 3: can you undo the bump, we bump from master. thanks.

andelf:
To reproduce the buf:

send:

```
packet 1:
058000002c800000c38000000000000000000000000a02c1f522086b0f67b0939b77f640a8ebb29e952e5acf01081f12ca010a31747970652e676f6f676c65617069732e636f6d2f70726f746f636f6c2e54726967676572536d617274436f6e74726163741294010a15416bf4248afac458049323aef8320d158f919c517212154147ea92fa0cfda6e02b7069c82d160f7f42073ced2264ed0813290000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000100000000000000000000000047ea92fa0cfda6e02b7069c82d160f7f42073ced

packet 2:
70a8b1af9e952e90018094ebdc03
```

After sending packet 1, expect to see a 0x9000 response and continue. But the device is alive and not responding.

PR #505, opened 2020-06-07
fix(TransportWebHID): device undefined error handling

hanjukim:
## Description
This fix handles uncaught exception `TypeError: Cannot read property 'productId' of undefined`. It only happens on Windows Chrome.

## Reproduce
On Windows Chrome,
1. start a new tab and try to access ledge using `hw-transport-webhid` module.
2. HID permission popup shows. Click cancel
3. exception will occur and observer does not pass it.

## The fix
Check for device existence, if it is not defined, call observer.error


gre:
Thanks!

PR #529, opened 2020-09-24
Introduce @ledgerhq/cryptoassets

gre:


.eslintignore
```diff
@@ -0,0 +1 @@
+packages/cryptoassets/data
```
meriadec, line 1: :rage1:

packages/cryptoassets/package.json
```diff
@@ -0,0 +1,37 @@
+{
+    "name": "@ledgerhq/cryptoassets",
+    "version": "5.23.0",
+    "description": "Ledger common errors",
```
meriadec, line 4: probably not the right description

.eslintignore
gre: yeah it's generated data, it's valid javascript. but eslint don't like `[,,,,stuff]` which is handy to save bytes.

packages/cryptoassets/package.json
gre: true

.eslintignore
meriadec: i was referring to the "missing EOL char" ^^ not so important though. usually it's better for git diff

PR #533, opened 2020-10-16
add polkadot currencies

henri-ly:
No need to merge now, still WIP


henri-ly:
nvm is it generated

gre:
@henri-ly actually it is not. this PR makes sense.

more generally speaking, **any coin** that is available in Manager as an app should gets in

PR #545, opened 2020-12-01
Prevent throw if device reset is failing

meriadec:
So, this PR mitigates impact of recent regression on Chrome 87 + Windows (reported at least 3 times on Vault side). The actual workaround is to disable "new WebUSB backend" (whatever that means) using chrome flag. With those changes no more need.

Tested on:
- Windows 10 + Chrome 87 (& validated I was falling in the error case + tested basic ADPUs)
- Linux + Chrome 87 (no apparent regression detected, which is not really surprising given the code lol)

What it look like (VM screen resolution FTW)

YAYA


gre (approved):
no problem for the codecov. there is no easy way to actually CI test this i think anyway.

PR #551, opened 2020-12-14
Fix v in transaction signatures returned from hw-app-eth.

jonored:
The APDU protocol for app-ethereum returns only the lowest byte of the v
value for the signature when signing EIP-155 transactions, relying on
the host-side code to provide the high bytes to achieve a valid
signature; this calculates and applies the prefix in hw-app-eth.


gre (approved):
Thanks for your help fixing this and adding tests too.

PR #570, opened 2021-02-08
hw-app-eth: v to be prefix by "0" to make a more value hex

gre:
on large chain ID blockchains:

before: v: "135"
after: v: "0135"


btchip (approved):
Tested locally for Matic with Metamask, no impact expected on other chains and integrations

PR #574, opened 2021-02-17
feat(cryptoassets): added registerCryptoCurrrency + some tests

rjarasson-ledger:
Cryptoassets package is used by live-common, but when implementing a new coin, we sometimes need to update the cryptoassets definition, and check that everything is ok - doing a yalc is a bit complicated sometimes in this case.

Also, when a 3rd party will need to add its own coin, it can be blocking to wait for his submission to be merged AND published in the cryptoassets package before being able to test anything.

I added a registerCryptoCurrency function to allow crypto currencies to be added in the list (and correctly handled) without the need to publish the package.

(this usage will be used in a mycoin integration example in the developer portal)


gre:
makes sense. thanks @rjarasson-ledger

PR #576, opened 2021-02-18
Fix EIP-155 chain id math

### LINKED ISSUE #575: Wrong v in signature when signing EIP-155 for high chain ids
When trying to use Ledger for signing with EWC on MyCrypto we were having issues with the returned `v` value in the signature. It looks like your math is slightly off for the high bits you append for high chain ids. 

https://github.com/LedgerHQ/ledgerjs/blob/master/packages/hw-app-eth/src/Eth.js#L211-L216

This should in fact be `CHAIN_ID * 2 + 35` as the base number for the high bits.

This looks to be working here at least:
https://github.com/MyCryptoHQ/MyCrypto/blob/57573908ae08f5097064b021cf3bc8a98e57d81a/src/services/WalletService/deterministic/ledger.ts#L75-L91

FrederikBolding:
Fixes #575

### DIFF
--- a/packages/hw-app-eth/src/Eth.js
+++ b/packages/hw-app-eth/src/Eth.js
@@ -208,7 +208,7 @@ export default class Eth {
       const chainIdSrc = rlpTx[6];
       const chainIdBuf = Buffer.alloc(4);
       chainIdSrc.copy(chainIdBuf, 4 - chainIdSrc.length);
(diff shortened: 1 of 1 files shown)


darkwing:
I can't wait to see this in -- I'll updated MetaMask with this change as soon as we see this published!

PR #587, opened 2021-03-31
[WIP] Add automated preparation of Ethereum device app for plugins and ERC20

TamtamHero:
This PoC PR adds a dependancy on `ethereum-tx-decoder`.
It also removes  `provideERC20TokenInformation` from the `decorateAppAPIMethods` list, to allow `signTransaction` to execute it as a subcall.

`signTransaction` is augmented to analyze the current transaction and send some preparation APDU to the device, such as `provideERC20TokenInformation` and/or `setExternalPlugin`.
The main advantage of doing this at the ledger.js level is to improve the support of device features on third party wallets like MEW/Metamask and to remove some manual logic from the Ledger Live to automate it.

It is possible to see it in action in this test: https://github.com/LedgerHQ/app-plugin-compound/blob/delegated_warning_ui/tests/src/compound_plugin.test.js 
You just need to link this custom `hw-app-eth` to the tests, run `dockerd`, and `Zemu` will do the rest.

A test version of ethereum is available on provider 4 (v1.6.5-rc1) but the compound plugin is not published yet.


packages/hw-app-eth/src/Eth.js
```diff
@@ -240,6 +249,22 @@ export default class Eth {
       toSend.push(buffer);
       offset += chunkSize;
     }
+
+    let decodedTx = decodeTx("0x" + rawTxHex);
+    if (decodedTx.data.length >= 10) {
+      const erc20Info = byContractAddress(decodedTx.to);
+      console.log(erc20Info);
+      if (erc20Info) {
+        this._provideERC20TokenInformation(erc20Info);
```
gre, line 258: in live-common, we also have to do this, specifically for compound to also provide the c* token data
https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/ethereum/signOperation.js#L72-L78

how do we plan to address this? would it still be needed?

machard:
also since the new game is to call provideERC20.... internally, we should update https://github.com/LedgerHQ/ledgerjs/pull/587/files#diff-2a96855bc41e2457f03a382c30057bb9257533493bad6acba9730305db50dd60R893 ?

machard:
still WIP: we need to handle automatically the multiple provideERC20Info automatically

packages/hw-app-eth/src/Eth.js
```diff
@@ -22,6 +22,9 @@ import { EthAppPleaseEnableContractData } from "@ledgerhq/errors";
 import type Transport from "@ledgerhq/hw-transport";
 import { BigNumber } from "bignumber.js";
 import { encode, decode } from "rlp";
```
machard, line 24: Since we switch to ethers, we should drop rlp and use ethers.utils.RLP.
However I tried and it's not a dropin replacement. Shoud be similar though.

packages/hw-app-eth/src/Eth.js
```diff
@@ -240,6 +234,35 @@ export default class Eth {
       toSend.push(buffer);
       offset += chunkSize;
     }
+
+    rlpTx = ethers.utils.RLP.decode("0x" + rawTxHex);
```
machard, line 238: this makes signTransactionLargeChainID2 test fail with "data array too short".
Should we catch silently this case or does the test needs to be fixed ?

machard: also just before this part, we modify rawTx in case its an EIP155 transaction.
Should we use the updated rawTx instead of the original rawTxHex?

TamtamHero: no idea for `signTransactionLargeChainID2`
But I'd say yes, you should use the updated rawTx here

packages/hw-app-eth/src/Eth.js
machard: the approve function selector is 0x095ea7b3
https://www.4byte.directory/signatures/?bytes4_signature=0x095ea7b3

we are going to update coumpound b2c json file to add this selector and set the erc20OfInterest key.
See paraswap https://github.com/LedgerHQ/crypto-assets/pull/204/files#diff-f1d134479e4f1cfc1ffe64248098a166ffe57fbc320bfb8dffdad190d10d2b70

machard:
- [ ] Rajouter approve de compound et les abi
- [ ] Importer need to handle the fact that selector without plugin set will not be in B2C_signature (so update the importer - - [ ] accordingly)
- [ ] Only call setExternalPlugin if plugin is set
- [ ] Do the to do
- [ ] Solve remaining questions in this pr

packages/hw-app-eth/src/Eth.js
machard: actually after discussion with @TamtamHero
- its USDT.approve(cUSDC) that is actually made as call.
- Instead of checking the ticker in the device app to know if its a coumpound approval they will hardcode directly the address

=> no more need to call proviceERC20TokenInformation

packages/hw-app-eth/src/Eth.js
machard: rawTx is actually not updated so its ok

packages/hw-app-eth/src/Eth.js
machard: => we need to leave the extra call in live common to ensure compatibility with previous ethereum app version

machard: https://github.com/LedgerHQ/app-ethereum/commit/43e7badfa4e0b4463f12483f556a048d6177b29f

packages/hw-app-eth/src/Eth.js
machard: done

gre:
we'll migrate in github actions in future because circleci do'nt run all commits somehow..

packages/hw-app-eth/src/Eth.js
machard: signTransactionLargeChainID2 fixed

machard:
need to import cal master branch when dapps json are merged

gre:
I'm merging this work on a branch called alpha so we can push a tagged version of the libraries. thanks for the work.

PR #590, opened 2021-04-12
add currencies for crypto.com

thomas-nguy:
Add support for crypto.com currencies


thomas-nguy:
[Wait for review by the team]

packages/cryptoassets/src/currencies.js
```diff
@@ -2698,6 +2726,36 @@ const cryptocurrenciesById: { [name: string]: CryptoCurrency } = {
       },
     ],
   },
+  crypto_org_croeseid: {
+    type: "CryptoCurrency",
+    id: "crypto_org_croeseid",
+    coinType: 1,
+    name: "Crypto.org Croeseid Coin",
+    managerAppName: "Crypto.org Croeseid Chain",
```
gre, line 2734: I don't see this application in my manager. Is it an advanced app?

thomas-nguy: Croeseid is the testnet for Crypto.org Chain.

I dont think there is any app in the manager for it. But I though there was also Bitcoin Test or Ethereum Ropsten which does not have an app?

Should I just remove it ? (I am actually not sure if it is supported)

gre: managerAppName must be a real app name. It sounds like it should be Crypto.org Chain, if you confirm you use that same app to also use testnet.
The same way ethereum ropsten currency have defined Ethereum for the app

thomas-nguy: Ok I see. I think it should be fine with "Crypto.org Chain" but I am not in the phase where I am able to test.
Nevertheless, I will change the value.

packages/cryptoassets/src/currencies.js
```diff
@@ -2698,6 +2726,36 @@ const cryptocurrenciesById: { [name: string]: CryptoCurrency } = {
       },
     ],
   },
+  crypto_org_croeseid: {
+    type: "CryptoCurrency",
+    id: "crypto_org_croeseid",
+    coinType: 1,
```
gre, line 2732: @TamtamHero what's your input about this: should the testnet accounts be derivated on same path as mainnet and therefore have same address or should it be on coinType 1?

existing choices:
- bitcoin will use 0 and 1
- ethereum will use 60 for both (because it's conventionally what everyone do, they derives testnet on 60' too)
present choice

TamtamHero: The device app only supports derivations on `m/44'/394'` so it doesn't make sense to add a testnet coinType here, nobody will be able to use it 
https://github.com/LedgerHQ/app-cryptocom/blob/cro/app/Makefile#L34

thomas-nguy: That is correct, the app only supports coin type 394.
I checked with the team, and it seems that even though it is specified to use 1' per spec, using 394' is still possible for croeseid addresses. Thanks I will modify the value accordingly

gre (approved):

PR #595, opened 2021-05-04
poc

valpinkman:


packages/devices/src/ble/receiveAPDU.ts
```diff
@@ -1,20 +1,15 @@
...
 import { log } from "@ledgerhq/logs";
-
 const TagId = 0x05;
-
 // operator that transform the input raw stream into one apdu response and finishes
 export const receiveAPDU = (
   rawStream: Observable<Buffer>
 ): Observable<Buffer> =>
-  Observable.create((o) => {
+  new Observable((o) => {
```
gre, line 9: why is it necessary?

gre (commented):
we'll need to try to be as conservative as possible regarding any JS change that our migration Flow->TypeScript do.

packages/devices/src/ble/receiveAPDU.ts
valpinkman: types say it's deprecated, also, this is just a PoC for now, nothing is settled yet

PR #596, opened 2021-05-04
LL-3928 Migrate to TypeScript

JunichiSugiura:
# Impact on end users of ledgerjs libraries

We designed the rework in a way that impact would be very low.

If you are using TypeScript, you should benefit the types of the library. If you were relying on FlowType: we have made a smooth transition curve where we would still support Flowtype defs for a few weeks (until we ourself fully migrate TypeScript in our projects) but we will eventually drop Flowtype completely (and make a major bump of ledgerjs)

The only type changes we did is on the Transport instance, we made it no longer have a parametric type,

so if you were writing code like

```
let transport: Transport<string> = Transport.open()
```

you can replace with

```
let transport: Transport = Transport.open()
```

---

We have also dropped the deprecated usage of 

```js
createPaymentTransactionNew(...multi params)
getWalletPublicKey(...multi params)
signP2SHTransaction(...multi params)
``` 

which should be replaced by the object style. It was console.warn as deprecated for 2 years so we expect it is not impacting much.

# TypeScript internal rework

## we used `flow-to-ts`
[GitHub - Khan/flow-to-ts: Convert flow code to typescript](https://github.com/Khan/flow-to-ts)
We have been using the codemod made by KhanAcademy to, more or less painlessly, convert our flow typed javascript files to typescript.
Most of the heavy lifting was done with the codemod, but we had to come back and fixes many types / errors.
For a quick usage of the codemod, here is how we used it
```bash
npx @khanacademy/flow-to-ts --prettier --write --delete-source --inline-utility-types src/**/*.js
```
We ran that script inside each `packages/*` folder (we decided to go modular and not run the script on the whole repo at once)

## Internal main changes
**./packages/cryptoassets/src/currencies.js**

```js

export type CryptoCurrencyObjMap<F> = $ObjMap<typeof cryptocurrenciesById, F>

export type CryptoCurrencyConfig<C> = CryptoCurrencyObjMap<(arg0: any) => C>

```

**./packages/hw-transport/src/Transport.js**

[TypeScript: Documentation - Classes](https://www.typescriptlang.org/docs/handbook/2/classes.html#type-parameters-in-static-members)
Cannot use Parametric Type in static members in Typescript.
Removed Descriptor in static members
Removed parametric type from Transport to make it simpler

**./packages/hw-app-algorand/src/Algorand.ts**

Added a `any` type to response in sign, as it can either be an object or an array from what I understand
```js
let response: any = null;
```

**./packages/hw-app-btc/src/Btc.ts**

Reworked without `arguments`  😱 and removing the deprecated calls
```js
getWalletPublicKey(
    path: string,
    opts?:
      | boolean
      | {
          verify?: boolean;
          format?: AddressFormat;
        }
  ): Promise<{
    publicKey: string;
    bitcoinAddress: string;
    chainCode: string;
  }> {
    let options;

    if (arguments.length > 2 || typeof opts === "boolean") {
      console.warn(
        "btc.getWalletPublicKey deprecated signature used. Please switch to getWalletPublicKey(path, { format, verify })"
      );
      options = {
        verify: !!opts,
        format: arguments[2] ? "p2sh" : "legacy",
      };
    } else {
      options = opts || {};
    }

    return getWalletPublicKey(this.transport, { ...options, path });
  }



createPaymentTransactionNew(arg: CreateTransactionArg) {
    if (arguments.length > 1) {
      console.warn(
        "@ledgerhq/hw-app-btc: createPaymentTransactionNew multi argument signature is deprecated. please switch to named parameters."
      );
      arg = fromDeprecateArguments(arguments, [
        "inputs",
        "associatedKeysets",
        "changePath",
        "outputScriptHex",
        "lockTime",
        "sigHashType",
        "segwit",
        "initialTimestamp",
        "additionals",
        "expiryHeight",
        "useTrustedInputForSegwit",
      ]);
    }

    return createTransaction(this.transport, arg);
  }


signP2SHTransaction(arg: SignP2SHTransactionArg) {
    if (arguments.length > 1) {
      console.warn(
        "@ledgerhq/hw-app-btc: signP2SHTransaction multi argument signature is deprecated. please switch to named parameters."
      );
      const [
        inputs,
        associatedKeysets,
        outputScriptHex,
        lockTime,
        sigHashType,
        segwit,
        transactionVersion,
      ] = arguments;
      arg = {
        inputs,
        associatedKeysets,
        outputScriptHex,
        lockTime,
        sigHashType,
        segwit,
        transactionVersion,
      };
      arg = fromDeprecateArguments(arguments, [
        "inputs",
        "associatedKeysets",
        "outputScriptHex",
        "lockTime",
        "sigHashType",
        "segwit",
        "transactionVersion",
      ]);
    }

    return signP2SHTransaction(this.transport, arg);
  }

```

**packages/hw-transport-node-ble/src/TransportNodeBle.ts**

I had to change the returnType because Promise<Buffer> would not work
Went for Promise<any> meanwhile
```js
exchange = (apdu: Buffer): Promise<Buffer> =>
    this.exchangeAtomicImpl(async () => {
      try {
        const msgIn = apdu.toString("hex");
        log("apdu", `=> ${msgIn}`);
        const data = await merge(
          this.notifyObservable.pipe(receiveAPDU),
          sendAPDU(this.write, apdu, this.mtuSize)
        ).toPromise();
        const msgOut = data.toString("hex");
        log("apdu", `<= ${msgOut}`);
        return data;
      } catch (e) {
        log("ble-error", "exchange got " + String(e));

        if (this.notYetDisconnected) {
          // in such case we will always disconnect because something is bad.
          await disconnectDevice(this.device).catch(() => {}); // but we ignore if disconnect worked.
        }

        throw e;
      }
    });

```

**packages/react-native-hid/package.json**
**pacjages/react-native-ble/package.json**

Removed the peerDependencies and moved them to dependencies because we did not find a solution to lint / build with peerDeps yet
- react-native@^0.63.4
- react-native-ble-plx@1.0.1

Apparently `rollup` can help us out with this issue.  We NEED to come back to this and fix it with peerDependencies

## Documentation
Remove the doc script from these breaking packages
- hw-transport-node-ble
- hw-transport-web-ble
- hw-transport-http
- hw-transport-mocker
- react-native-hw-transport-ble

## Tests
Changed the way BTC tests were written because they were still using the old API. Fixed the test to reflect the deprecation of the old APIs


valpinkman:
# TS Rework
## flow-to-ts
[GitHub - Khan/flow-to-ts: Convert flow code to typescript](https://github.com/Khan/flow-to-ts)
We have been using the codemod made by KhanAcademy to, more or less painlessly, convert our flow typed javascript files to typescript.
Most of the heavy lifting was done with the codemod, but we had to come back and fixes many types / errors.
For a quick usage of the codemod, here is how we used it
```bash
npx @khanacademy/flow-to-ts --prettier --write --delete-source --inline-utility-types src/**/*.js
```
We ran that script inside each `packages/*` folder (we decided to go modular and not run the script on the whole repo at once)


## Important Changes
**./packages/cryptoassets/src/currencies.js**

```js

export type CryptoCurrencyObjMap<F> = $ObjMap<typeof cryptocurrenciesById, F>

export type CryptoCurrencyConfig<C> = CryptoCurrencyObjMap<(arg0: any) => C>

```

**./packages/hw-transport/src/Transport.js**

[TypeScript: Documentation - Classes](https://www.typescriptlang.org/docs/handbook/2/classes.html#type-parameters-in-static-members)
Cannot use Parametric Type in static members in Typescript.
Removed Descriptor in static members
Removed parametric type from Transport to make it simpler

**./packages/hw-app-algorand/src/Algorand.ts**

Added a `any` type to response in sign, as it can either be an object or an array from what I understand
```js
let response: any = null;
```

**./packages/hw-app-btc/src/Btc.ts**

Reworked without `arguments`  😱 and removing the deprecated calls
```js
getWalletPublicKey(
    path: string,
    opts?:
      | boolean
      | {
          verify?: boolean;
          format?: AddressFormat;
        }
  ): Promise<{
    publicKey: string;
    bitcoinAddress: string;
    chainCode: string;
  }> {
    let options;

    if (arguments.length > 2 || typeof opts === "boolean") {
      console.warn(
        "btc.getWalletPublicKey deprecated signature used. Please switch to getWalletPublicKey(path, { format, verify })"
      );
      options = {
        verify: !!opts,
        format: arguments[2] ? "p2sh" : "legacy",
      };
    } else {
      options = opts || {};
    }

    return getWalletPublicKey(this.transport, { ...options, path });
  }



createPaymentTransactionNew(arg: CreateTransactionArg) {
    if (arguments.length > 1) {
      console.warn(
        "@ledgerhq/hw-app-btc: createPaymentTransactionNew multi argument signature is deprecated. please switch to named parameters."
      );
      arg = fromDeprecateArguments(arguments, [
        "inputs",
        "associatedKeysets",
        "changePath",
        "outputScriptHex",
        "lockTime",
        "sigHashType",
        "segwit",
        "initialTimestamp",
        "additionals",
        "expiryHeight",
        "useTrustedInputForSegwit",
      ]);
    }

    return createTransaction(this.transport, arg);
  }


signP2SHTransaction(arg: SignP2SHTransactionArg) {
    if (arguments.length > 1) {
      console.warn(
        "@ledgerhq/hw-app-btc: signP2SHTransaction multi argument signature is deprecated. please switch to named parameters."
      );
      const [
        inputs,
        associatedKeysets,
        outputScriptHex,
        lockTime,
        sigHashType,
        segwit,
        transactionVersion,
      ] = arguments;
      arg = {
        inputs,
        associatedKeysets,
        outputScriptHex,
        lockTime,
        sigHashType,
        segwit,
        transactionVersion,
      };
      arg = fromDeprecateArguments(arguments, [
        "inputs",
        "associatedKeysets",
        "outputScriptHex",
        "lockTime",
        "sigHashType",
        "segwit",
        "transactionVersion",
      ]);
    }

    return signP2SHTransaction(this.transport, arg);
  }

```

**packages/hw-transport-node-ble/src/TransportNodeBle.ts**

I had to change the returnType because Promise<Buffer> would not work
Went for Promise<any> meanwhile
```js
exchange = (apdu: Buffer): Promise<Buffer> =>
    this.exchangeAtomicImpl(async () => {
      try {
        const msgIn = apdu.toString("hex");
        log("apdu", `=> ${msgIn}`);
        const data = await merge(
          this.notifyObservable.pipe(receiveAPDU),
          sendAPDU(this.write, apdu, this.mtuSize)
        ).toPromise();
        const msgOut = data.toString("hex");
        log("apdu", `<= ${msgOut}`);
        return data;
      } catch (e) {
        log("ble-error", "exchange got " + String(e));

        if (this.notYetDisconnected) {
          // in such case we will always disconnect because something is bad.
          await disconnectDevice(this.device).catch(() => {}); // but we ignore if disconnect worked.
        }

        throw e;
      }
    });

```

**packages/react-native-hid/package.json**
**pacjages/react-native-ble/package.json**

Removed the peerDependencies and moved them to dependencies because we did not find a solution to lint / build with peerDeps yet
- react-native@^0.63.4
- react-native-ble-plx@1.0.1

Apparently `rollup` can help us out with this issue.  We NEED to come back to this and fix it with peerDependencies

packages/hw-transport-node-hid-noevents/src/TransportNodeHid.ts
```diff
@@ -15,27 +13,26 @@ import type { DeviceModel } from "@ledgerhq/devices";
...
 const filterInterface = (device) =>
-  ["win32", "darwin"].includes(process.platform)
-    ? // $FlowFixMe
-      device.usagePage === 0xffa0
+  ["win32", "darwin"].includes(process.platform) // $FlowFixMe
+    ? device.usagePage === 0xffa0
     : device.interface === 0;
 
-export function getDevices(): Array<*> {
+export function getDevices(): Array<any> {
```
Sajjon, line 20: We should try to avoid using `any` at all costs, should we not return some kind of `Device` type here? I thought this was `Device` of [`@types/node-hid`](https://www.npmjs.com/package/@types/node-hid)? 

```typescript
export interface Device {
    vendorId: number;
    productId: number;
    path?: string;
    serialNumber?: string;
    manufacturer?: string;
    product?: string;
    release: number;
    interface: number;
    usagePage?: number;
    usage?: number;
}
```

Or some similar type?

packages/hw-transport-node-hid-noevents/src/TransportNodeHid.ts
```diff
@@ -45,14 +42,12 @@ export default class TransportNodeHidNoEvents extends Transport<?string> {
...
-  static list = (): Promise<(?string)[]> =>
+  static list = (): Promise<any> =>
     Promise.resolve(getDevices().map((d) => d.path));
 
   /**
    */
-  static listen = (
-    observer: Observer<DescriptorEvent<?string>>
-  ): Subscription => {
+  static listen = (observer: Observer<DescriptorEvent<any>>): Subscription => {
```
Sajjon, line 50: Is `Subscription` an `rxjs` Subsciption? hard to tell since it is imported from `hw-transport` and not rxjs.

Anyway, I think this method is kind of awkward, I think it better if the method returned an `Observable<Event>` that you subscribe to yourself, instead of accepting an observer as arg and returning a subscription. The reason for this is that if we were to return an Observable we can pipe it! And do stuff with it, whereas a Subscription is.... well end of the road really.

Also the `DescriptorEvent<any>` should probably be massaged a bit. Better if it is a tagged union without the `any` generic?

Maybe:
```typescript
type LedgerAvailabilityEvent = LedgerAvailabilityEventAvailable | LedgerAvailabilityEventUnavailable

type LedgerAvailabilityEventAvailable = Readonly<{
    type: 'new'
    // Device that was recently found
   device: Device
}>

type LedgerAvailabilityEventUnavailable = Readonly<{
    type: 'lost'
    // Device that is not available anymore
   device: Device
}>
```

Or similar.

packages/hw-transport-node-hid-noevents/src/TransportNodeHid.ts
```diff
@@ -144,22 +144,22 @@ export default class TransportNodeHidNoEvents extends Transport<?string> {
    * @param apdu
    * @returns a promise of apdu response
    */
-  exchange = (apdu: Buffer): Promise<Buffer> =>
+  exchange = (apdu: Buffer): Promise<any> =>
```
Sajjon, line 147: Why the change from `Promise<Buffer>`  to `Promise<any>`?

packages/hw-transport-node-hid-noevents/src/TransportNodeHid.ts
gre: I agree with your feedback, we should take note for future but right now we are porting code without changing the interface.
To me it's still discussable: It was initially meant to be loosely coupled with rxjs: someone that depends on a hw-transport-* do not bring the rxjs package in with the whole bundle. That way, you can even still use this without rxjs.

agreed for union type.

valpinkman:
Merge the work of https://github.com/LedgerHQ/ledgerjs/pull/602 into this PR

JunichiSugiura:
@gre I added a temporary module called `ledger-support` which has `yarn copy` script to copy all the `.flow.js` file from the latest `@ledgerhq/*` packages. Installation is automatically done right after running `yarn` on ledgerjs project root and `(cd flow-support && yarn copy)` is called when running `yarn build`. We need not to forget setting the latest flow versions in `flow-support/package.json` explicitly just before publishing TS version of ledgerjs.

JunichiSugiura:
We need to find a way to deal with Transport changes in old `.js.flow`.

gre:
@JunichiSugiura what is the transport changes? 🤔  we are supposed to not change anything. if we do, it's a breaking and we shouldn't break.
if it's about the parametric changes, we don't have to deal with this as we don't want to change the live-common side neither regarding fixing flowtype to remove parametirc type.

gre:
@JunichiSugiura what do you think is the sunset strategy for "flowsupport" ? is it something we want to keep forever? 6 months sunset? etc.. we have to announce it in the release notes to me.

JunichiSugiura:
It's depends on how often this repo introduces breaking changes. unless we edit '.flow.js' manually from now on when interfaces changes or new mod is introduced.

gre:
PR replaced by https://github.com/LedgerHQ/ledgerjs/pull/619

Issue #607, opened 2021-05-27
Problem with WebUSB + Chrome 91.X


humanumbrella:
Mostly an FYi as I am still searching for answers.

After updating to Chrome 91, when attempting to connect to either Ledger device (X or S), I am seeing the opportunity to pair the device as normal, what looks like a successful pairing, but then when I try to do any operation, I am getting `The device was disconnected.`

This is for existing applications that work fine on Chrome 90.X.

humanumbrella:
Device shows up and appears to be behaving properly in chrome://usb-internals/

Ultimately, I am trying to use `const transport = await TransportWebUsb.create()` - and now this does not work in newest Chromium version or any browser that uses it. It does not seem to matter if I am connecting via a saved (previously paired) device or have reset that and connecting fresh. I get the same outcome. Pasting some output from chrome://device-log/

```
USBUser[13:50:56] USB device added: vendor=11415 "Ledger", product=4117 "Nano S", serial="0001", guid=02b2cba5-69d7-4984-814f-40f524495323

USBEvent[13:50:56] Failed to open device: No such device (it may have been disconnected)

USBUser[13:50:56] USB device removed: guid=7d99e9c9-ff3b-47b4-9281-5a0d0dd14294

USBEvent[13:50:56] Failed to claim interface: Other error

USBUser[13:49:21] USB device added: vendor=11415 "Ledger", product=4117 "Nano S", serial="0001", guid=7d99e9c9-ff3b-47b4-9281-5a0d0dd14294

USBEvent[13:49:20] Failed to open device: No such device (it may have been disconnected)

USBUser[13:49:20] USB device removed: guid=197e5832-548f-43c0-bc7e-b9294581db7b

USBEvent[13:49:20] Failed to claim interface: Other error
```

Any clients that have Chrome auto-update turned on are now affected and unable to use their devices. Please advise.

cc @meriadec as you fixed `hw-transport-webusb` last time.

humanumbrella:
I didn't mention but I already tried to flip the bit in chrome://flags to `NEW USB Backend` to **Disabled** 

It still does not work, the behavior is:

First try - `Unable to claim interface.`
Second try - `The device was disconnected.`

humanumbrella:
Bump - this is a major issue, please advise. I am continuing to debug further. 

Is anyone from Ledger able to assist?

MattG561:
[...]

PR #609, opened 2021-06-03
Removes call to reset after call to open

humanumbrella:
An attempt to fix issue https://github.com/LedgerHQ/ledgerjs/issues/607


gre (commented):
Thanks for this iteration. We started the same exploration yesterday on our end and we are validating the solution today.

I was wondering why you kept it in `close()`?
I think we'll have to kill the reset usage completely.

gre (approved):
We did assess that it indeed is fixing the problem so it can be merged.
we are still trying to understand if we also will need to keep it for close(). for now we keep it and we will test more in depth with the next release.

PR #611, opened 2021-06-04
added equilibrium currency

alex-nax:


gre:
Can you add more context to this? Is there an application available in the Ledger Nano devices? we aim to only track the currencies that are at least supported in our device. thanks.

alex-nax:
Equilibrium is a substrate-based chain. You can read about us [here](https://equilibrium.io/en)
Recently we did swap DOT<->EQ as part of a PLO phase, some of our clients transfered DOTs from their ledger wallet, so we are building a ledger app that they can claim their EQ tokens, it is still work in progress, you can find it [here](https://github.com/equilibrium-eosdt/ledger-equilibrium)

gre:
we can keep this PR for reference, we can reopen later when the app is about to get published. NB that the coin type don't appear on https://github.com/satoshilabs/slips/blob/master/slip-0044.md so i think we'll need to make sure the coinType is correct in future;
For now, as PR can quickly become conflicting with master, it's better to close and reopen as soon as things are getting ready on the app side.

Thanks

alex-nax:
As for now EQ coin type was added to slip-0044 https://github.com/satoshilabs/slips/pull/1019 a long time ago.
By the way we've applied to Ledger Listing Program at the end of June and still got no response

PR #619, opened 2021-06-09
LL-3928 TypeScript migration

gre:
# Impact on end users of ledgerjs libraries

We designed the rework in a way that impact would be very low.

If you are using TypeScript, you should benefit the types of the library. If you were relying on FlowType: we have made a smooth transition curve where we would still support Flowtype defs for a few weeks (until we ourself fully migrate TypeScript in our projects) but we will eventually drop Flowtype completely (and make a major bump of ledgerjs)

The only type changes we did is on the Transport instance, we made it no longer have a parametric type,

so if you were writing code like

```
let transport: Transport<string> = Transport.open()
```

you can replace with

```
let transport: Transport = Transport.open()
```

---

We have also dropped the deprecated usage of 

```js
createPaymentTransactionNew(...multi params)
getWalletPublicKey(...multi params)
signP2SHTransaction(...multi params)
``` 

which should be replaced by the object style. It was console.warn as deprecated for 2 years so we expect it is not impacting much.

# TypeScript internal rework

## we used `flow-to-ts`
[GitHub - Khan/flow-to-ts: Convert flow code to typescript](https://github.com/Khan/flow-to-ts)
We have been using the codemod made by KhanAcademy to, more or less painlessly, convert our flow typed javascript files to typescript.
Most of the heavy lifting was done with the codemod, but we had to come back and fixes many types / errors.
For a quick usage of the codemod, here is how we used it
```bash
npx @khanacademy/flow-to-ts --prettier --write --delete-source --inline-utility-types src/**/*.js
```
We ran that script inside each `packages/*` folder (we decided to go modular and not run the script on the whole repo at once)

## Internal main changes
**./packages/cryptoassets/src/currencies.js**

```js

export type CryptoCurrencyObjMap<F> = $ObjMap<typeof cryptocurrenciesById, F>

export type CryptoCurrencyConfig<C> = CryptoCurrencyObjMap<(arg0: any) => C>

```

**./packages/hw-transport/src/Transport.js**

[TypeScript: Documentation - Classes](https://www.typescriptlang.org/docs/handbook/2/classes.html#type-parameters-in-static-members)
Cannot use Parametric Type in static members in Typescript.
Removed Descriptor in static members
Removed parametric type from Transport to make it simpler

**./packages/hw-app-algorand/src/Algorand.ts**

Added a `any` type to response in sign, as it can either be an object or an array from what I understand
```js
let response: any = null;
```

**./packages/hw-app-btc/src/Btc.ts**

[...]

PR #620, opened 2021-06-09
LL-4600 Map HID errors to DisconnectedDeviceDuringOperation

juan-cortes:
Following the discussion over at https://ledgerhq.atlassian.net/browse/LL-4600, we want to introduce better wording for the case where `node-hid` cannot write to the device. Currently, since it's not being mapped, we are getting the error directly from https://github.com/node-hid/node-hid/blob/master/src/HID.cc#L413 and we need to throw our own custom error to introduce translations and specific wording.

I dropped the `isDisconnectedError` in favor of a remap but perhaps there's something else to be done here, I don't like relying on the error messages themselves but I see no other alternative when I look at the source for node-hid. This would still need specific wording on LLD to take advantage of the remap, as far as mobile goes I'm not sure it applies, but I could be wrong.

### DIFF
--- a/packages/hw-transport-node-hid-noevents/src/TransportNodeHid.ts
+++ b/packages/hw-transport-node-hid-noevents/src/TransportNodeHid.ts
@@ -10,7 +10,11 @@ import { ledgerUSBVendorId } from "@ledgerhq/devices";
 import hidFraming from "@ledgerhq/devices/lib/hid-framing";
 import { identifyUSBProductId, identifyProductName } from "@ledgerhq/devices";
 import { DeviceModel } from "@ledgerhq/devices";
-import { TransportError, DisconnectedDevice } from "@ledgerhq/errors";
+import {
+  TransportError,
+  DisconnectedDevice,
+  DisconnectedDeviceDuringOperation,
+} from "@ledgerhq/errors";
 
 const filterInterface = (device) =>
   ["win32", "darwin"].includes(process.platform)
@@ -21,8 +25,6 @@ export function getDevices(): (Device & { deviceName?: string })[] {
   return HID.devices(ledgerUSBVendorId, 0x0).filter(filterInterface);
 }
 
-const isDisconnectedError = (e) =>
-  e && e.message && e.message.indexOf("HID") >= 0;
 /**
  * node-hid Transport minimal implementation
  * @example
@@ -30,6 +32,14 @@ const isDisconnectedError = (e) =>
  * ...
  * TransportNodeHid.create().then(transport => ...)
  */
+
+// Mapping errors from https://github.com/node-hid/node-hid/blob/master/src/HID.cc
+const mapError = (e) => {
+  if (e && e.message) {
+    return new DisconnectedDeviceDuringOperation(e);
+  }
+  return e;
+};
 
 export default class TransportNodeHidNoEvents extends Transport {
   /**
@@ -109,12 +119,12 @@ export default class TransportNodeHidNoEvents extends Transport {
       this.device.write(data);
       return Promise.resolve();
     } catch (e) {
-      if (isDisconnectedError(e)) {
+      const mappedError = mapError(e);
+      if (mappedError instanceof DisconnectedDeviceDuringOperation) {
         this.setDisconnected();
-        return Promise.reject(new DisconnectedDevice(e.message));
       }
 
-      return Promise.reject(e);
+      return Promise.reject(mappedError);
     }
   };
   readHID = (): Promise<Buffer> =>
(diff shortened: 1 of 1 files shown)


packages/hw-transport-node-hid-noevents/src/TransportNodeHid.js
```diff
@@ -25,8 +29,20 @@ export function getDevices(): Array<*> {
...
+// Mapping errors from https://github.com/node-hid/node-hid/blob/master/src/HID.cc
+// verbatim inconsistent casing.
+const mapError = (e) => {
+  if (e && e.message) {
+    switch (e.message) {
+      case "Cannot write to closed device":
+      case "could not read from HID device":
+        return new DisconnectedDevice(e);
+      case "Cannot write to hid device":
+        return new CantWriteDevice(e);
```
gre, line 41: I have feeling that we don't need these two error at all and could stick to the

DisconnectedDeviceDuringOperation

Which is exactly that paradigm of "unexpected interruption" which is the case for many transport implementation too.

juan-cortes: You probably know more about this than I do, the remapping was merely to cover the "Cannot write to hid device" error message from reaching the user

gre: essentially i suggest we remap it all to DisconnectedDeviceDuringOperation
let's check if this idea is acceptible regarding the wording it produces but imo all "DisconnectedDeviceDuringOperation" means "you interrupted something when we were trying to exchange with the device"

juan-cortes: Do we have a way to reproduce the error consistently?

gre: i guess unplug during an apdu exchange, but not sure. probably another way to glitch usb.

gre:
@juan-cortes can you resolve conflicts? since merge of https://github.com/LedgerHQ/ledgerjs/pull/619 :sorry:

packages/hw-transport-node-hid-noevents/src/TransportNodeHid.ts
```diff
@@ -31,6 +33,14 @@ const isDisconnectedError = (e) =>
  * TransportNodeHid.create().then(transport => ...)
  */
 
+// Mapping errors from https://github.com/node-hid/node-hid/blob/master/src/HID.cc
+const mapError = (e) => {
```
gre, line 37: i think we can totally get rid of this method because it looks like it's always going to be that remap?

packages/hw-transport-node-hid-noevents/src/TransportNodeHid.ts
```diff
@@ -109,12 +119,12 @@ export default class TransportNodeHidNoEvents extends Transport {
       this.device.write(data);
       return Promise.resolve();
     } catch (e) {
-      if (isDisconnectedError(e)) {
+      const mappedError = mapError(e);
+      if (mappedError instanceof DisconnectedDeviceDuringOperation) {
```
gre, line 123: same here?

gre (changes requested):
trying to understand something, we can remap literally everything now, right?

packages/hw-transport-node-hid-noevents/src/TransportNodeHid.ts
```diff
@@ -125,12 +128,16 @@ export default class TransportNodeHidNoEvents extends Transport {
...
+            e && e.message
+              ? new DisconnectedDeviceDuringOperation(e.message)
+              : e;
+          if (maybeMappedError instanceof DisconnectedDeviceDuringOperation) {
             this.setDisconnected();
             return reject(new DisconnectedDevice(e.message));
           }
 
-          reject(e);
+          return Promise.reject(maybeMappedError);
```
gre, line 140:
```suggestion
          reject(maybeMappedError);
```

packages/hw-transport-node-hid-noevents/src/TransportNodeHid.ts
```diff
@@ -125,12 +128,16 @@ export default class TransportNodeHidNoEvents extends Transport {
...
 
         if (e) {
-          if (isDisconnectedError(e)) {
+          const maybeMappedError =
+            e && e.message
+              ? new DisconnectedDeviceDuringOperation(e.message)
+              : e;
+          if (maybeMappedError instanceof DisconnectedDeviceDuringOperation) {
             this.setDisconnected();
             return reject(new DisconnectedDevice(e.message));
```
gre, line 137:
```suggestion
```

gre (changes requested):

gre:
work here need to be fixed and fully tested with LLD.

PR #621, opened 2021-06-11
Expose a way to get Model from targetId

juan-cortes:
This all comes from the context of wanting to expose the model id in a nice way for the proxy transport instead of relying on a fallback to nanoS like we do for `listApps` https://github.com/LedgerHQ/ledger-live-common/blob/master/src/apps/hw.js#L124 we could use the `deviceInfo` we already have access to and enhance it with the device model data. Or even modify the signature of `DeviceInfo` to include the model data at some point.

This allows us to unlock LNX proxied LLM development, especially in tasks involving the manager where this nanoS fallback brings storage problems.

### DIFF
--- a/packages/devices/README.md
+++ b/packages/devices/README.md
@@ -16,11 +16,13 @@ Logic for all Ledger devices.
 -   [ledgerUSBVendorId](#ledgerusbvendorid)
 -   [getDeviceModel](#getdevicemodel)
     -   [Parameters](#parameters-1)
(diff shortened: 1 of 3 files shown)


gre (approved):

gre (changes requested):
please regenerate documentation for CI to pass.

juan-cortes:
> please regenerate documentation for CI to pass.

`yarn doc` right? committed the changes

PR #622, opened 2021-06-22
NEAR currency info

volovyk-s:
Please, answer questions in the comments and I will update this PR.

### DIFF
--- a/packages/cryptoassets/src/currencies.ts
+++ b/packages/cryptoassets/src/currencies.ts
@@ -78,6 +78,25 @@ const ethereumUnits = (name, code) => [
 ];
 
 const cryptocurrenciesById: Record<string, CryptoCurrency> = {
+  near: {
+    type: "CryptoCurrency",
+    id: "near",
+    coinType: 0, //TODO: what should I put here?
+    name: "NEAR", //TODO: is it the name of the company, or name of the cryptocurrency?
+    managerAppName: "NEAR",
+    ticker: "N",
+    scheme: "near", //TODO: where this will be used? `near` is ok?
+    color: "#000", //TODO: where this collor woll be used?
+    family: "near", //TODO: where this will be used? `near` is ok?
+    units: [
+      {
+        name: "N",
+        code: "N",
+        magnitude: 8, //TODO: what is the magnitude? How is it used?
+      },
+    ],
+    explorerViews: [],
+  },
   aeternity: {
     type: "CryptoCurrency",
     id: "aeternity",
--- a/packages/cryptoassets/src/currencies.ts
+++ b/packages/cryptoassets/src/currencies.ts
@@ -81,18 +81,18 @@ const cryptocurrenciesById: Record<string, CryptoCurrency> = {
   near: {
     type: "CryptoCurrency",
     id: "near",
-    coinType: 0, //TODO: what should I put here?
-    name: "NEAR", //TODO: is it the name of the company, or name of the cryptocurrency?
+    coinType: 397,
+    name: "NEAR",
     managerAppName: "NEAR",
-    ticker: "N",
-    scheme: "near", //TODO: where this will be used? `near` is ok?
-    color: "#000", //TODO: where this collor woll be used?
-    family: "near", //TODO: where this will be used? `near` is ok?
+    ticker: "NEAR",
+    scheme: "near",
+    color: "#000",
+    family: "near",
     units: [
       {
-        name: "N",
-        code: "N",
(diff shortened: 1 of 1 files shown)


gre:
Is there a nano app planned? Adding currency in this library only happens once an app is available in manager (at least in staging)

volovyk-s:
@gre yes, NEAR app is already available in production: https://github.com/LedgerHQ/app-near
Nafissa asked me in slack to add NEAR to this list.

packages/cryptoassets/src/currencies.ts
```diff
@@ -78,6 +78,25 @@ const ethereumUnits = (name, code) => [
 ];
 
 const cryptocurrenciesById: Record<string, CryptoCurrency> = {
+  near: {
+    type: "CryptoCurrency",
+    id: "near",
+    coinType: 0, //TODO: what should I put here?
+    name: "NEAR", //TODO: is it the name of the company, or name of the cryptocurrency?
```
gre, line 85: name of crypto currency.

packages/cryptoassets/src/currencies.ts
```diff
@@ -78,6 +78,25 @@ const ethereumUnits = (name, code) => [
 ];
 
 const cryptocurrenciesById: Record<string, CryptoCurrency> = {
+  near: {
+    type: "CryptoCurrency",
+    id: "near",
+    coinType: 0, //TODO: what should I put here?
```
gre, line 84: this usually is the coinType as in BIP44 / SLIPS44.
it looks like you are 397 in https://github.com/satoshilabs/slips/blob/master/slip-0044.md

packages/cryptoassets/src/currencies.ts
```diff
@@ -78,6 +78,25 @@ const ethereumUnits = (name, code) => [
...
+    name: "NEAR", //TODO: is it the name of the company, or name of the cryptocurrency?
+    managerAppName: "NEAR",
+    ticker: "N",
```
gre, line 87: mmh I think the ticker used in markets are `NEAR`, no? https://www.coingecko.com/fr/pi%C3%A8ces/near#markets

packages/cryptoassets/src/currencies.ts
```diff
@@ -78,6 +78,25 @@ const ethereumUnits = (name, code) => [
...
+    scheme: "near", //TODO: where this will be used? `near` is ok?
```
gre, line 88: typically it could allow (in future) a `near://` uri scheme. we could make this optional in future, you can leave it as "near"

packages/cryptoassets/src/currencies.ts
```diff
@@ -78,6 +78,25 @@ const ethereumUnits = (name, code) => [
...
+    color: "#000", //TODO: where this collor woll be used?
```
gre, line 89: this usually is the main color you have defined in the app logo.

packages/cryptoassets/src/currencies.ts
```diff
@@ -78,6 +78,25 @@ const ethereumUnits = (name, code) => [
...
+    family: "near", //TODO: where this will be used? `near` is ok?
```
gre, line 90: yeah, it is == to the id you have chosen, so let it be "near".

packages/cryptoassets/src/currencies.ts
```diff
@@ -78,6 +78,25 @@ const ethereumUnits = (name, code) => [
...
+    units: [
+      {
+        name: "N",
+        code: "N",
```
gre, line 94: I think same, to me this should be NEAR.

packages/cryptoassets/src/currencies.ts
```diff
@@ -78,6 +78,25 @@ const ethereumUnits = (name, code) => [
...
+        magnitude: 8, //TODO: what is the magnitude? How is it used?
```
gre, line 95: this is very important to get it right when we plan an integration: it represent the pow of 10 of the most minimal unit of the currency.

For instance, in Bitcoin, it's 8 because 1 BTC = 10^8 satoshi.
as we would use "integers" to represent the smallest divisible amount, it needs to be representing this.

for Ethereum as an other example, it's 18.

gre (changes requested):

packages/cryptoassets/src/currencies.ts
volovyk-s: Ok, the smallest one is yN (yocto NEAR). So it should be 24.

volovyk-s:
@gre thanks! I have updated this PR, please, review.

PR #635, opened 2021-07-29
Add axios to hw-app-eth dependencies

avsetsin:
This PR adds `axios` to `hw-app-eth` dependencies. It was added in the last release and is used here: https://github.com/LedgerHQ/ledgerjs/blob/master/packages/hw-app-eth/src/contracts.ts#L1


gre:
good catch. thanks.

PR #636, opened 2021-07-30
Workaround fix for null chainID if EIP 155 marker intersects packet boundary

hectorchu:
The Ledger Eth app appears to exhibit a bug whereby if the EIP 155 marker (containing the chainID and null r,s values) is split at the APDU boundary, then the app fails to parse the chainID correctly and sets it to 0. This is most easily observed when the chainID is 4 bytes (as with Harmony ONE).

This workaround avoids the split by enlarging the chunk to incorporate the rest of the data if a split would happen. Since the EIP 155 marker is only expected to be 7 bytes or less, this does not increase the chunk size far beyond the 150 byte soft maximum.

### DIFF
--- a/packages/hw-app-eth/src/Eth.ts
+++ b/packages/hw-app-eth/src/Eth.ts
@@ -239,9 +239,9 @@ export default class Eth {
           ? rawTx.length - offset
           : maxChunkSize;
 
-      if (rlpOffset != 0 && offset + chunkSize == rlpOffset) {
+      if (rlpOffset != 0 && offset + chunkSize >= rlpOffset) {
         // Make sure that the chunk doesn't end right on the EIP 155 marker if set
-        chunkSize--;
+        chunkSize = rawTx.length - offset;
       }
 
       const buffer = Buffer.alloc(


gre:
do you have a transaction test example that would be ok to publicly add in a public test? thanks.

hectorchu:
Sure, this is a transaction that exhibits the problem:
`f9011781f985012a05f200830370de9415527b33bef39f5deb8aa0360aed18ad9982e5d888016345785d8a0000b8e47ff36ab50000000000000000000000000000000000000000000000000037278bf5fcaa00000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000061068a8a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000cf664087a5bb0237a0bad6742852ec6c8d69a27a000000000000000000000000f315803ba9da293765ab163e7db98e8d6df6d3618463564c408080`

hectorchu:
What do I need to do to get this merged?

pscott:
> What do I need to do to get this merged?

We need the tests to successfully pass :)
Currently working on it :)

pscott:
@hectorchu Thanks for your PR.

So I thought about the "Since the EIP 155 marker is only expected to be 7 bytes or less" argument. Unless I'm mistaken, I think `chainID` could be any `int256` value. Which means it could go as big as `32` bytes.
However, given that the APDU size max is `0xff`, even in the worst case scenario, we won't be hitting this limit.

So I think your commit looks good. I'm currently fixing the tests because this commit "breaks" some tests (since is introduces some change in the way APDUs are generated).

hectorchu:
Ok great. I was only saying that I expect 32 bits to be all that we need in the way of chain IDs, we won't have so many different chains that we will exceed this.

pscott:
If you look at the recent merge on master, you will see that we just allowed for bigger chain IDs :)

Indeed, some people are going for BIG chain IDs… and we can’t tell them not to do it unfortunately :(

PR #637, opened 2021-07-31
Add EIP1559 and EIP2930 support

FrederikBolding:
Small changes required to support the new transaction formats for the new transaction types.

To be merged with changes here: https://github.com/LedgerHQ/app-ethereum/pull/157


FrederikBolding:
Thanks @pscott for help with the APDUs 😄

packages/hw-app-eth/src/Eth.ts
```diff
@@ -262,12 +267,28 @@ export default class Eth {
       offset += chunkSize;
     }
 
-    rlpTx = ethers.utils.RLP.decode("0x" + rawTxHex);
```
gre, line 265: @machard : do you know why we needed to prefix 0x in the past? Also do you know why we had to call it twice. I wonder if we could use a higher level function that would have avoided having to do the logic of that pr on our side. Because ethers must have somewhere the logic to extract the rlp part of a tx?

gre (approved):
Thanks for contribution!

Internal note: It's curious that we were not future proof via usage of rlp  but I feel we use a too low level utility from it. 🤔

In any case this should be good to go, I asked internally what was the reason of the previous 0x prefix but as test are passing here it should be fine 🙂

gre:
I'm going to merge this work as it seems to pass CI. thanks a lot for also adding some tests for this scenario, very useful.
I will issue a ledgerjs version straightaway so we can already anticipate the incoming app.

PR #641, opened 2021-08-06
Allow for big chainID

pscott:
This PR allows the usage of chainIDs bigger than 4 bytes.
It does so by recovering the the truncated chainID that gets computed by the device itself (the device does its computation using the 4 lowest bytes, not the full chainID).

This will be removed later on, when the device gets proper support, and when proper support for big chain ID are also added to Ledger (most notably for erc20 tokens).


pscott:
Closes #640

packages/hw-app-eth/src/Eth.ts
```diff
@@ -202,9 +202,10 @@ export default class Eth {
     );
 
     let rlpOffset = 0;
-    let chainIdPrefix = "";
+    let chainId = 0;
```
gre, line 205: this must be `new BigNumber(0)` because if you don't enter the if below, `(0).time(2)` is going to badly break.

TypeScript spotted it as the CI is failing right now.

gre (changes requested):

packages/hw-app-eth/src/Eth.ts
pscott: done :)

gre (changes requested):
we'll need to have test to cover this. let's work on this together next week.

gre (approved):

PR #650, opened 2021-08-29
Add json-rpc package (replaces legacy web3-subprovider)

jsonjasono:
Figured this was a good starting point to replacing the legacy web3-subprovider package, as using the raw transports is difficult without quite a bit of experience.


votelessdan (approved):

gre:
the former web3-subprovider package was out of ledgerjs because hard to maintain so would be this package. That said, it's a pretty good idea but we don't have yet the maturity to have this granularity of modularisation, at the moment we tend to have such logic embedded in our Ledger Live stack – anyone is free to have this code on their side & I hope later we figure out this loosely couple way to make it back as a standalone OSS package.

Thanks.

PR #654, opened 2021-09-02
Fix an issue with ERC20 info when using EIP 1559 transactions

FrederikBolding:
This should fix an issue where ERC20 information wasn't provided to the Ledger when using EIP 1559 transactions!

Looks to work, but might need a sanity check 😄


packages/hw-app-eth/src/Eth.ts
```diff
@@ -197,14 +197,40 @@ export default class Eth {
...
+        data: rlpDecoded[6],
+        to: rlpDecoded[4],
+        chainId: rlpTx[0],
+      };
+    } else {
+      // Legacy tx
+      decodedTx = {
+        data: rlpDecoded[5],
+        to: rlpDecoded[3],
+        chainId: rlpTx[6],
```
pscott, line 230: `rlpTx[6]` is only valid for post EIP-155 transactions (see https://eips.ethereum.org/EIPS/eip-155)
If `rlpTx.length < 6` then the `chainId` should default to `1` :)

pscott (changes requested):
LGTM! Moving the `decodedTx` up and putting the `chainId` in it is indeed a proper way to deal with this.

Left a single request for change :D

packages/hw-app-eth/src/Eth.ts
FrederikBolding: Will change 👍

PR #656, opened 2021-09-10
Support PlatON(LAT)

zfamz:
if it's ok. Can you update the @ledgerhq/cryptoassets pack. and publish the @ledgerhq/hw-app-lat pack ^ _ ^


gre:
The codebase of this PR appears to fork a lot those from Ethereum app, which doesn't make sense to me as I don't think it map exactly as-is. I miss the context too.

zfamz:
> The codebase of this PR appears to fork a lot those from Ethereum app, which doesn't make sense to me as I don't think it map exactly as-is. I miss the context too, but if it's a ETH fork, it might make more sense to keep being in hw-app-eth, the same way BSC today can work with hw-app-eth.

The PlatON project is a public chain of the POS protocol modified by fork Ethereum. The external interface part maintains compatibility with Ethereum, so most of the external applications of Ethereum can be transplanted to the PlatON chain.

PR #658, opened 2021-09-16
LL-6794 Add ability to overrides plugins + be chainId ready

gre:
- added a way to disable the baseURL fetching of plugin
- added a way to explicitely provide the plugins data instead of fetching it
- [x] add test to cover the new mechanism
  - [x] disable fully the baseURL should disable plugin
  - [x] disable the baseURL + hardcoded plugin config should enable plugin 
- [ ] ~~update the "importer" to split per chainId too~~ can't do this until this happens in CAL
  
  
## gist

```js
const eth = new Eth(transport);
eth.setPluginsLoadConfig({
  baseURL: null, // disable fully the data fetching
  extraPlugins: { ...some plugins... } // specify extra plugins to be added
})
eth.setPluginsLoadConfig({
  baseURL:"https://cdn.live.ledger-stg.com", // use a staging api
})
```


achabrolin-ledger (approved):

machard (approved):

gre:
i'll rollback the idea to split per chainId because there is possibly no need to do so.

The idea i had was:

```
LEGACY: https://cdn.live.ledger.com/plugins/ethereum.json
NEW: https://cdn.live.ledger.com/plugins/ethereum/${chainId}.json
```

pscott:
Late to the party but thanks for the work @gre  :)

PR #663, opened 2021-09-27
LL-6672  - Add NFT related errors

ghost:
Adding 2 new errors for the send of NFTs.

```javascript
export const NotOwnedNft = createCustomErrorClass("NotOwnedNft");
```
To throw when trying to send an NFT not synced in the account.

<hr/>

```javascript
export const NotEnoughNftOwned = createCustomErrorClass("NotEnoughNftOwned");
```
To throw when trying to send too many ERC1155 NFTs compared to the amount owned by the account.


gre:
please add them in https://github.com/LedgerHQ/ledger-live-common/blob/master/src/errors.ts instead because it's not domain of ledgerjs errors

PR #666, opened 2021-10-06
getWalletPublicKey supports new btc app

kallerosenbaum:
This adds support for the new btc app. Currently, only Btc.getWalletPublicKey() should work for the new app. If the old app is used, nothing should have changed, functionality-wise, with this PR.

We make this intermediate PR so that we can make an internal release of ledgerjs to be tested by QA, while we keep working on Btc.createPaymentTransactionNew().


gre (commented):
very clean code overall. great to also see some more lib utility being used. some comments

ghost (approved):

PR #669, opened 2021-10-07
LL-8000 – Add support for NFTs (721 and 1155)

pscott:
This PR adds support for NFTs to the `signTransaction` function.

`provideForContract` now calls `provideNFTInformation` before calling `provideERC20Information`.
`provideNFTInformation` uses `getNFTInformation` (defined in `nfts.ts`). This function calls ledger's backend to query signed information regarding the NFT the tx is potentially interacting with.

The same logic is applied to `loadInfosForContractMethod` (i.e. first check if the contract is an NFT, if not then check if it's an ERC20).


Taweesakpintakam7m1 (approved):

packages/hw-app-eth/src/contracts.ts
```diff
@@ -36,21 +36,13 @@ export const loadInfosForContractMethod = async (
...
-          if (
-            e.response &&
-            400 <= e.response.status &&
-            e.response.status < 500
-          ) {
-            return null; // not found cases can be ignored to allow future changes in endpoint without failing a signature to be done.
-          }
-          throw e;
-        });
+        .catch(() => null);
```
gre, line 45: not a good idea to hide all possible error cases. but maybe there is a reason?

packages/hw-app-eth/src/nfts.ts
```diff
@@ -0,0 +1,51 @@
...
+  contractAddress: string;
+  collectionName: string;
+  data: Buffer;
+};
+
+type BackendResponse = {
+  payload: string;
+};
+
+const BACKEND_URL = "https://nft.staging.aws.ledger.fr/v1/ethereum";
```
gre, line 13: - we need to dev a way to change that URL programmatically/env (i can take it)
- we need a stable endpoint

gre (commented):
remaining topics

- CI to fix
- backend to give us a production endpoint
- need to add new tests to cover the new features

packages/hw-app-eth/src/contracts.ts
pscott: Since this API will be used by everyone (including external wallets), if we throw we will just cancel the transaction. The idea here to not make our backend become a single point of failure for all external wallets ^^

gre: to be defined what shall be the behavior but it could be a "network is down" or a "undefined is not a function" (e.G. bug in axios) and in these 2 cases, the error should be thrown because it make a flawky support of a feature 🤔

gre: i'll keep your change and add a log

gre:
^ will push small changes to implement these.

gre (approved):

PR #698, opened 2021-11-10
Fix sequence=0 becomes 0xffffffff when signing

kallerosenbaum:
Also fixed two other non-broken but ugly instances of "if (number | undefined)"


kallerosenbaum:
Fixes https://ledgerhq.atlassian.net/jira/software/c/projects/LL/issues/LL-8092

kallerosenbaum:
Fixes https://ledgerhq.atlassian.net/browse/LL-7946

hzheng-ledger (approved):
looks good to me. Thanks

PR #700, opened 2021-11-11
update viacoin explorer

romanornr:
Change Viacoin old block explorer based on insight to blockbook


FabriceDautriat:
This PR will not solve the broadcast topic as it only relates to the block explorer used to check transaction details only after the tx is successfully broadcasted.

Ledger uses its own explorers and nodes to broadcast ViaCoin transactions.

romanornr:
Ok, understood

At this moment Ledger uses fail to transfer their funds
Including Binance

Can Ledger check if the status of their own explorer/node and see if the status is okay or down

If there are issues, let me know if there are details and what I can do about it.

Issue #701, opened 2021-11-11
support for new transaction format in peercoin


backpacker69:
Recently peercoin has been updated to new transaction format, from version 3 onwards serialized transaction does not have a separate timestamp field, that was originally there after version. as a result, signing input from these new transactions fails, as deserialisation expects timestamp to be present:

packages/hw-app-btc/src/splitTransaction.ts:

```
 38   if (hasTimestamp) {
 39     timestamp = transaction.slice(offset, 4 + offset);
 40     offset += 4;
 41   }
```

I propose to add handling of this to splitTransaction, by adding "peercoin" to additionals when calling createPaymentTransactionNew method and checking for transaction versions 1 and 2 before attempting to extract timestamp.

Also, communication part in app-bitcoin would have to be changed to something like this:

+++ b/src/btchip_transaction.c
@@ -268,8 +268,9 @@ void transaction_parse(unsigned char parseMode) {
                     }

                     if (G_coin_config->flags & FLAG_PEERCOIN_SUPPORT) {
-                        if ((G_coin_config->family ==
-                            BTCHIP_FAMILY_PEERCOIN) ||
+                        if (((G_coin_config->family ==
+                            BTCHIP_FAMILY_PEERCOIN) &&
+                            (btchip_context_D.transactionVersion[0] < 3)) ||
                             ((G_coin_config->family == BTCHIP_FAMILY_STEALTH) &&
                             (btchip_context_D.transactionVersion[0] < 2))) {
                             // Timestamp

PR #702, opened 2021-11-11
Peercoin v11 and new explorer

backpacker69:
related to https://github.com/LedgerHQ/ledgerjs/issues/701

### DIFF
--- a/packages/cryptoassets/src/currencies.ts
+++ b/packages/cryptoassets/src/currencies.ts
@@ -1749,8 +1749,8 @@ const cryptocurrenciesById: Record<string, CryptoCurrency> = {
     ],
     explorerViews: [
       {
-        tx: "https://explorer.peercoin.net/tx/$hash",
-        address: "https://explorer.peercoin.net/address/$address",
+        tx: "https://blockbook.peercoin.net/tx/$hash",
+        address: "https://blockbook.peercoin.net/address/$address",
       },
     ],
   },
--- a/packages/hw-app-btc/src/signP2SHTransaction.ts
+++ b/packages/hw-app-btc/src/signP2SHTransaction.ts
@@ -31,6 +31,7 @@ export type SignP2SHTransactionArg = {
   sigHashType?: number;
   segwit?: boolean;
   transactionVersion?: number;
+  initialTimestamp?: number;
 };
 export async function signP2SHTransaction(
   transport: Transport,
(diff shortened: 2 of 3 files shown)


gre:
So if i'm correct this fix will only work with the same fix done on the nano app at https://github.com/LedgerHQ/app-bitcoin/pull/210 , correct?

backpacker69:
this fix will work alone insofar that ledgerjs library will be able to split new peercoin transactions properly without crashing in the middle, but fix LedgerHQ/app-bitcoin#210 is also necessary for end users of ledger live to be able to sign v3+ peercoin transactions and actually be capable of spending peercoins from v3+ inputs.

what i couldn't find is where additionals are populated in ledger live, so this line would work and set this variable properly.

`const isPeercoin = additionals.includes("peercoin");
`

FabriceDautriat:
Linked to : https://github.com/LedgerHQ/app-bitcoin-new/pull/10
@tjulien-ledger

gre (changes requested):
instead of introducing one more flag, we are considering having less intelligence in ledgerjs but more on user land, which means , instead of the suggested changes, we prefer to let the user land set `hasTimestamp` to false when it satisfy the blockchain(s).

we have ongoing test that confirm this approach works https://github.com/LedgerHQ/ledger-live-common/pull/1621/files

as for the explorersViews change, feel free to send a dedicated PR for it but it's better to split the two topics. thanks

backpacker69:
> https://github.com/LedgerHQ/ledger-live-common/pull/1621/files

problem with this approach is that it cannot handle inputs in older version transactions, which means that you won't be able to spend old coins in your wallet for example. I cannot figure out how to control this from user land, set hasTimestamp to false before calling to split  but how to call sign transaction with both types?

hzheng-ledger:
@backpacker69 
good point.
I think I should do this:
`        if (version !== "01000000" && version !== "02000000") {
          hasTimestamp = false;
        }
        else{
          hasTimestamp = true;
        }

`
So, "hasTimestamp = true" for inputs in older version, "hasTimestamp = false" for inputs in new version.
Do I understand correctly?

backpacker69:
I think that if hasTimestamp parameter can be reconfigured according to transaction version, then it should work.

hzheng-ledger:
Yes, I agree.
Also, I see that you suggest to change the peercoin explorer url from "https://explorer.peercoin.net" to "https://blockbook.peercoin.net"
Are they different?

peerchemist:
> Yes, I agree. Also, I see that you suggest to change the peercoin explorer url from "https://explorer.peercoin.net" to "https://blockbook.peercoin.net" Are they different?

Both lead to the same service, but we are trying to migrate to only "blockbook.peercoin.net" domain.

gre:
My understand is we managed to fix it on the "user land" side so the logic can happen outside of the library and this fix is no longer needed. Don't hesitate to comment if i'm wrong. Thanks.

PR #724, opened 2021-11-24
add helium asset info to cryptoassets

allenan:
adds asset info for Helium (HNT) to the `@ledgerhq/cryptoassets` package

### DIFF
--- a/packages/cryptoassets/src/currencies.ts
+++ b/packages/cryptoassets/src/currencies.ts
@@ -1209,6 +1209,35 @@ const cryptocurrenciesById: Record<string, CryptoCurrency> = {
     ],
     explorerViews: [],
   },
+  helium: {
+    type: "CryptoCurrency",
+    id: "helium",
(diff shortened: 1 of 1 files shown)


packages/cryptoassets/src/currencies.ts
```diff
@@ -1209,6 +1209,35 @@ const cryptocurrenciesById: Record<string, CryptoCurrency> = {
     ],
     explorerViews: [],
   },
+  helium: {
+    type: "CryptoCurrency",
+    id: "helium",
+    coinType: 904,
+    name: "Helium",
+    managerAppName: "Helium",
```
gre, line 1217: I don't see any app called Helium at the moment in the Manager. cc @FabriceDautriat

FabriceDautriat: There should be one. Maybe only visible in developer mode.

gre (approved):

Issue #725, opened 2021-11-24
Decentraland MANA incorrectly delisted and countervalue disabled


markymark617:
It seems that commit [349dcaa19199781221549dbcee68ffc9b24ee026](https://github.com/LedgerHQ/ledgerjs/commit/349dcaa19199781221549dbcee68ffc9b24ee026#diff-eb12ef28b59b70ab7ab47380126774a14f0866c91d07f0ff95a1101c876f5fc4) delisted MANA by setting `delisted` to `true` and set `disableCounterValue` to `true` also.

Both of these used to be `false`. I will create a bugfix branch with this fix and raise a PR for approval.

PR #726, opened 2021-11-24
updating `disableCountervalue` and `delisted` to FALSE for decentraland MANA

markymark617:
Resolves open issue #725


gre:
this file is generated from our internal database but i'll inform the team to do this so it gets back into a future PR. thanks

PR #737, opened 2021-12-09
LL-8068: Improve the performance of the BLE data transfers

sportron-ledger:
Tested on android so far, need to be tested with ios as well.
Two changes for BLE transfer boosting:
- Put the BLE connection priority to high by default to reduce the connection interval.
- Use write without response to send apdus if such characteristic exists, this new characteristic we'll be provided soon by LNX 2.0.2-rc1 firmware version.


juan-cortes (approved):
🚀 lgtm

gre (approved):

PR #746, opened 2021-12-27
Add Speculos Http API transport

r4mmer:
I just added a package similar to speculos transport but using the http api instead of the tcp ports


gre (changes requested):
please add a README.md and run `yarn doc`

r4mmer:
> please add a README.md and run `yarn doc`

Done!

gre (approved):
thanks

PR #747, opened 2022-01-03
Rework hw-app-eth to split out "ledger services" from the app logic

gre:
With this PR,

```ts
eth.signTransaction(path, txHex)
```

becomes **deprecated** in favor of

```ts
eth.signTransaction(path, txHex, resolution)
```

where resolution is an object with following shape:

```ts
{
  erc20Tokens: Array<string>;
  nfts: Array<string>;
  externalPlugin: Array<{ payload: string; signature: string }>;
  plugin: Array<string>;
}
```

This object contains all the hex serialized data the device may need to **clear sign** information on device screen.

**Hopefully we also provide a default "service"** to resolve this data, example:

```ts
import ledgerService from "@ledgerhq/hw-app-eth/lib/services/ledger"
async function main() {
  ...
  const resolution = await ledgerService.resolveTransaction(txHex)
    .catch(() => null); // <-- up to you if you want a fallback to blind sign. (which was previous behavior)
   const result = eth.signTransaction("44'/60'/0'/0/0", txHex, resolution);
}
```

This allows us to decouple the transaction resolution from the device signature logic itself in order for you to have better control and for more usecases
- an alternative service (rather than Ledger's) could be used.
- we plan to make even more things dynamic, typically the way we load and resolve ERC20 and plugins are too monolithic. This rework pave the path to allow this even more easily.
- you may want to load the resolve data way ahead in your "validation process" and typically inform the user if an error occurs before it reach the device phase.
- you may want a finer control weither you allow blind sign or not. Typically, you may or may not use `.catch(() => null)` in the snippet above


packages/hw-app-eth/src/Eth.ts
```diff
@@ -305,137 +271,42 @@ export default class Eth {
...
-          chainIdTruncated,
-          this.loadConfig
-        );
+    const response_byte: number = response[0];
+    let v = "";
 
-        if (infos) {
-          const { plugin, payload, signature, erc20OfInterest, abi } = infos;
+    if (chainId.times(2).plus(35).plus(1).isGreaterThan(255)) {
+      const oneByteChainId = (chainIdTruncated * 2 + 35) % 256;
```
haammar-ledger, line 287: Fully trusting you on the buffer voodoo here and below :stuck_out_tongue:

ghost (approved):

haammar-ledger:
No added test coverage for the new signature?

packages/hw-app-eth/src/Eth.ts
gre: I only moved code on this one.

PR #748, opened 2022-01-03
LL-8827 Introduce types libraries

gre:
This PR introduces 3 new libraries.

- `@ledgerhq/types-live` (Account type, Operation, …)
- `@ledgerhq/types-cryptoassets` (Currency, TokenCurrency,..)
- `@ledgerhq/types-devices` (Transport, Device,...)

in order to do more decoupling between our types and the actual implementations and in a future goal to dedup our libraries.

https://ledgerhq.atlassian.net/browse/LL-8827


gre:
One possible idea, is to have some subtyping like this:

```
export type BitcoinAccount = Account & { bitcoinResources: BitcoinResources }
```

and all parts where we would need it, we would need to cast,

```
const bitcoinAccount = account as BitcoinAccount;
```

and that BitcoinAccount type would be part of a coin integration family specific indeed.

juan-cortes:
What is wrong with your proposed idea? I find it elegant and explicit enough 😕

haammar-ledger:
I don't see any issue either with that proposal.

gre:
@juan-cortes @haammar-ledger ok, then I can check my checkbox :)

PR #755, opened 2022-01-06
LL-8906 implement better resilience of hw-app-eth on resolution service

gre:


packages/hw-app-eth/src/Eth.ts
```diff
@@ -213,7 +213,10 @@ export default class Eth {
           " + const resolution = await ledgerService.resolveTransaction(rawTxHex);"
       );
       resolution = await ledgerService
-        .resolveTransaction(rawTxHex, this.loadConfig)
+        .resolveTransaction(rawTxHex, this.loadConfig, {
+          externalPlugins: true,
+          erc20: true,
+        })
```
gre, line 219: preserve the previous default behavior which wasn't resolving the nfts.

packages/hw-app-eth/src/services/ledger/index.ts
```diff
@@ -12,7 +12,7 @@ import { getNFTInfo, loadNftPlugin } from "./nfts";
 import { decodeTxInfo } from "../../utils";
 
 const ledgerService: LedgerEthTransactionService = {
-  resolveTransaction: async (rawTxHex, loadConfig) => {
+  resolveTransaction: async (rawTxHex, loadConfig, resolutionConfig) => {
```
gre, line 15: now, with the resolutionConfig, which is basically user's intent on things to resolve, we can branch the resolution of precisely what is really needed.

ghost (approved):

PR #760, opened 2022-01-11
Fix wrong bytestring parsing in NFT informations

lambertkevin:
Fixing the wrong parsing inside of the Eth app. 
No impact at all since it's purely for logging purposes.

Spec is here: https://ledgerhq.atlassian.net/wiki/spaces/WALLETCO/pages/3269984297/NFT-1+NFT+Backend+design#NFT-Metadata-BLOB


grsoares21 (approved):

packages/hw-app-eth/src/services/ledger/nfts.ts
```diff
@@ -32,9 +32,11 @@ export const getNFTInfo = async (
     .catch(axiosErrorHandling);
   if (!response) return;
 
+  // APDU response specification: https://ledgerhq.atlassian.net/wiki/spaces/WALLETCO/pages/3269984297/NFT-1+NFT+Backend+design#NFT-Metadata-BLOB
   const payload = response["payload"];
-  const collectionNameLength = Number(payload.slice(2, 3));
-  const collectionName = payload.slice(3, 3 + collectionNameLength).toString();
+  // Collection name length position: 3rd byte -> caracter 4 to 6
+  const collectionNameLength = parseInt(payload.slice(4, 6), 16);
+  const collectionName = payload.slice(6, 6 + collectionNameLength).toString();
```
grsoares21, line 39: is collectionNameLength in number of bytes? shouldn't we multiply by 2 the collectionNameLength in this case?

grsoares21 (changes requested):

packages/hw-app-eth/src/services/ledger/nfts.ts
lambertkevin: Yep wrong version of my PR 🤦

PR #778, opened 2022-01-27
LL-9033 Handle inner exception after pairing timeout

juan-cortes:
## 🦒 Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LL-9033

## 💻  Description 
This is a bit of a pickle, for context the new `fw 2.0.2` introduced some changes to the new characteristic, the one that brought the performance boost `13d63400-2c97-0004-0003-4c6564676572`, which threw an error from inside the [monitor characteristic](https://github.com/dotintent/react-native-ble-plx/blob/c8af4dba2ec02a822b6ef3bb4199244498b88326/android/src/main/java/com/polidea/reactnativeble/BleClientManager.java#L700) call from ble-plx.

We have to note that this is **only when we reach the 30s timeout when the pairing code is shown.** You need to be pairing a new device, ignore the pairing prompt, and reach this crash. If you feel this change introduces a performance hit (I measured and looked heavily into the `instanceof` perf and it seems pretty _light_) please advice for a better solution. I banged my head for two days and this is the only solution I could find that doesn't totally silence the error or hinder the functionality.

As an explanation of why this is happening, my understanding is that when we [**create the observable outside of the transport instance**](https://github.com/LedgerHQ/ledgerjs/blob/39ff487e4fc84b84b49fb0d6a07d20feb904e7a7/packages/react-native-hw-transport-ble/src/BleTransport.ts#L213-L218) or the open method, the exception thrown is unhandled and therefor breaks the application. By [**catching that specific error**](https://github.com/LedgerHQ/ledgerjs/blob/a7830322d2be4d1cafd08a865ec26eea7595547c/packages/react-native-hw-transport-ble/src/BleTransport.ts#L222-L228) in the observable, channeling it into the `inferMTU` method, and rethrowing it from there, we get the handled error and the correct "Pairing Failed" screen instead of an app crash. 

In other words, without this, the error is unhandled.

### DIFF
--- a/packages/react-native-hw-transport-ble/src/BleTransport.ts
+++ b/packages/react-native-hw-transport-ble/src/BleTransport.ts
@@ -13,12 +13,20 @@ import type { DeviceModel } from "@ledgerhq/devices";
 import { sendAPDU } from "@ledgerhq/devices/lib/ble/sendAPDU";
 import { receiveAPDU } from "@ledgerhq/devices/lib/ble/receiveAPDU";
 import { log } from "@ledgerhq/logs";
-import { Observable, defer, merge, from } from "rxjs";
-import { share, ignoreElements, first, map, tap } from "rxjs/operators";
+import { Observable, defer, merge, from, of, throwError } from "rxjs";
+import {
+  share,
+  ignoreElements,
+  first,
+  map,
+  tap,
+  catchError,
+} from "rxjs/operators";
 import {
   CantOpenDevice,
   TransportError,
   DisconnectedDeviceDuringOperation,
+  PairingFailed,
 } from "@ledgerhq/errors";
 import type { Device, Characteristic } from "./types";
 import { monitorCharacteristic } from "./monitorCharacteristic";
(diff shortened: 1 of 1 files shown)


packages/react-native-hw-transport-ble/src/BleTransport.ts
```diff
@@ -445,6 +461,9 @@ export default class BluetoothTransport extends Transport {
         mtu =
           (await merge(
             this.notifyObservable.pipe(
+              tap((maybeError) => {
+                if (maybeError instanceof Error) throw maybeError;
```
gre, line 465: are you sure throwing from a `tap` works? It feels a bit weird but good to know if it does, it makes the underlying observable failing?

that said, this part really feels weird to me, like why would an event of the notify be an instance of Error, we are making the type "any" here which isn't great either. maybe it's a `Buffer|PairingFailed` to be more specific?

gre (changes requested):

packages/react-native-hw-transport-ble/src/BleTransport.ts
juan-cortes: I'm sure it works (the tap throw) because I can handle the error that was crashing the app.
It all **seems weird to me too!** yes, but I was 100% unable to catch the error otherwise, the error is thrown inside the monitor method, so inside the ble-plx, and if we don't handle it this way. Ie, if we throw from outside the transport itself, then it's unhandled on the UI and crashes the app.

As for the typing I couldn't get it to like exactly that `Buffer|PairingFailed` but if you think it can work, I'm open to suggestions.

juan-cortes:
Think we can merge this until we figure out a cleaner solution @gre it addresses two bugs for 2.0.2 
I don't think we can solve the typing issue without introducing checking overhead to make TS happy, we'd have to check that it's a Buffer and not an error every time.

https://ledgerhq.atlassian.net/browse/LIVE-1400
https://ledgerhq.atlassian.net/browse/LIVE-579

gre (approved):

PR #780, opened 2022-01-28
LL 8968 - Add strict mode to prevent blind signing in live

lambertkevin:
This PR add the notion of `strictMode` for the `Eth` class in order to throw errors on specific cases related to transaction resolution.

[LL-8968](https://ledgerhq.atlassian.net/browse/LL-8968)

### DIFF
--- a/packages/hw-app-eth/src/Eth.ts
+++ b/packages/hw-app-eth/src/Eth.ts
@@ -20,6 +20,7 @@ import { BigNumber } from "bignumber.js";
 import { decodeTxInfo } from "./utils";
 // NB: these are temporary import for the deprecated fallback mechanism
 import { LedgerEthTransactionResolution, LoadConfig } from "./services/types";
+import { defaultLoadConfig } from "./services/ledger/loadConfig";
 import ledgerService from "./services/ledger";
 import {
   EthAppNftNotSupported,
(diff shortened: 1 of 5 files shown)


packages/hw-app-eth/src/Eth.ts
```diff
@@ -1201,43 +1236,52 @@ function provideNFTInformation(
...
     () => true,
     (e) => {
-      if (e && e.statusCode === 0x6a80) {
+      if (e?.statusCode === 0x6a80) {
         // this case happen when the plugin name is too short or too long
-        return false;
-      } else if (e && e.statusCode === 0x6984) {
+        return strictModeHandler(strictMode, e);
+      } else if (e?.statusCode === 0x6984) {
         // this case happen when the plugin requested is not installed on the device
```
gre, line 1254: in this case, there is a possibility, even inside Ledger Live, that you enter this case and it's actually "legit". Example: contract that accept funds in and would be plugin. it's a edge case, but it's a possibility: users trying to do a regular send eth to a contract and didn't installed the possible plugin that relates to it would be forced to install it. in that case we don't seem to offer a clear error so i don't think we want to do it like this, we probably want to have specific error because the case is "legit" to not have installed the plugin and we may want to drive to a specific wording / clear error, on this case.
we need to catch up with someone knowing deeply the app and its edge cases.

gre (changes requested):

PR #783, opened 2022-01-31
LL-8968 throw error of provideNFTInformation

gre:
this implement the minimal part of LL-8968 that is important for the NFT feature.


gre:
confirmed to work as expected. (manual test)

PR #787, opened 2022-01-31
chore: add missing @helium/proto dependency

elbywan:
Package `@ledgerhq/hw-app-helium` added with [this PR](https://github.com/LedgerHQ/ledgerjs/pull/768) has a missing dependency. 

`@helium/proto` itself is technically a dependency of other `@helium/…` packages but it is not good practice to directly import transitive dependencies like this (and the build will fail with `pnpm` since it is stricter than `yarn`).

```
src/Helium.ts(12,19): error TS2307: Cannot find module '@helium/proto' or its corresponding type declarations.
```

Capture d’écran 2022-01-31 à 14 58 02
Capture d’écran 2022-01-31 à 14 58 09

### DIFF
--- a/packages/hw-app-helium/package.json
+++ b/packages/hw-app-helium/package.json
@@ -28,6 +28,7 @@
     "license": "Apache-2.0",
     "dependencies": {
         "@helium/crypto": "^3.45.0",
+        "@helium/proto": "^1.5.0",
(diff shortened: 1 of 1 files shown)


gre:
it's curious why our CI didn't see it. thanks.

PR #797, opened 2022-02-08
LL-8941 Conditionally auto disconnect the node-hid-singleton

juan-cortes:
Here's the second attempt at getting a automatic release of the node-hid-singleton. After the talk we had the other day @gre I think that by exposing this method `setAllowAutoDisconnect` at the transport level and allowing it to override the auto-disconnect (basically re-scheduling it again) will allow us to call `transport.setAllowAutoDisconnect(false)` at the begining of the `withDevice` implementation on `hw/deviceAccess` from [live-common](https://github.com/LedgerHQ/ledger-live-common/blob/master/src/hw/deviceAccess.ts#L108) and unsetting the flag on the `onClose` for the transport, we essentially bypass the problem we have of long running tasks inside a `withDevice` instance.

What do you think?

### DIFF
--- a/packages/hw-transport-node-hid-singleton/src/TransportNodeHid.ts
+++ b/packages/hw-transport-node-hid-singleton/src/TransportNodeHid.ts
@@ -12,6 +12,23 @@ import { identifyUSBProductId } from "@ledgerhq/devices";
 import { CantOpenDevice } from "@ledgerhq/errors";
 import { listenDevices } from "./listenDevices";
 let transportInstance;
+
+const DISCONNECT_TIMEOUT = 5000;
+let disconnectTimeout;
+const clearDisconnectTimeout = () => {
+  if (disconnectTimeout) {
+    clearTimeout(disconnectTimeout);
+  }
+};
+
+const setDisconnectTimeout = () => {
+  clearDisconnectTimeout();
+  disconnectTimeout = setTimeout(
+    () => TransportNodeHidSingleton.autoDisconnect(),
+    DISCONNECT_TIMEOUT
+  );
+};
+
 /**
  * node-hid Transport implementation
  * @example
@@ -21,6 +38,7 @@ let transportInstance;
  */
 
 export default class TransportNodeHidSingleton extends TransportNodeHidNoEvents {
+  preventAutoDisconnect = false;
   /**
    *
    */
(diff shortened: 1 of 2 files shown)


packages/hw-transport-node-hid-singleton/src/TransportNodeHid.ts
```diff
@@ -88,6 +105,19 @@ export default class TransportNodeHidSingleton extends TransportNodeHidNoEvents
     };
   };
 
+  /**
+   * convenience wrapper for auto-disconnect logic
+   */
+  static async autoDisconnect(): void {
```
gre, line 111: can we call this function just `disconnect()` ? some transport have it and it's the same idea.

packages/hw-transport-node-hid-singleton/src/TransportNodeHid.ts
```diff
@@ -12,6 +12,22 @@ import { identifyUSBProductId } from "@ledgerhq/devices";
...
+const DISCONNECT_TIMEOUT = 5000;
+let disconnectTimeout;
+const clearDisconnectTimeout = () => {
+  if (disconnectTimeout) {
+    clearTimeout(disconnectTimeout);
+  }
+};
+
+const setDisconnectTimeout = () => {
+  disconnectTimeout = setTimeout(
```
gre, line 25: shouldn't it implicitly clearDisconnectTimeout here in case it was called twice?

gre (commented):
idea lgtm, some feedback

packages/hw-transport-node-hid-singleton/src/TransportNodeHid.ts
juan-cortes: You're probably right, I'll check :beau:

juan-cortes:
I would like a tagged version of the `node-hid-singleton` package so that I can make a PR on live-common that disables auto-disconnect inside `withDevice` and then test it on LLD. Think we can release this to test @gre ? I'd rather not do it considering the mess I caused last time 🙈

gre:
- can you regen doc?
- i'll soon issue a new ledgerjs bump from master, after i do can you also merge in the latest master?
- then yeah i can do a custom release (which is documented in README)

juan-cortes:
I think this is good to merge again @gre ?

PR #800, opened 2022-02-17
Add PLC Ultima (plcu) support

plcudeveloper:


gre:
Thanks for the PR. we can use it as reference in the future, but for now an nano app need to exist first before it gets in ledgerjs library.

PR #808, opened 2022-03-09
20220309 CAL update

adrienlacombe:


ghost (approved):
LGTM but I don't know what to check.
@gre Is there anything specific to check this kind of update, and when should we merge them?

adrienlacombe-ledger:
I think you would need to do a "release" of the cryptoassets package, if that makes sense and then include this new version everywhere. I let @gre confirm. Tahn kyou :pray:

gre:
ledgerjs crypto list = if it pass the ci it's good to merge.

now for the Ledger Live side: the ledgerjs package just was released a day ago with previous iteration, so i guess this will need to go in the next future train we have on this stack

adrienlacombe-ledger:
yes np, thank you @gre :)

PR #822, opened 2022-03-28
LIVE-1173: remove old btt trc10

henri-ly:
https://ledgerhq.atlassian.net/browse/LIVE-1173

Remove BTT and rename it (token 1002000) from the add token list
also fix the script of import token and


script/crypto-assets-importer/tron/trc10-tokens.js
```diff
@@ -1,5 +1,5 @@
 var whitelist = [
-  1002000, // BTT
+  // 1002000, // BTT
```
haammar-ledger, line 2: Is that comment expected, shouldn't it be removed?
(Genuine question, I have no idea how it's supposed to work. :smile: )

ghost (commented):
Questions about the changes in the `trc10.js` file:
- Why the change of IDs from string to number?
- It doesn't seem the BTT token is removed from the file? (2 different instances btw... I'm not even sure which one should be removed)
- Seems there's an addition at the end of a "TRON"/"TRX" token, is this expected?

henrily-ledger:
- It's seems that we used to have r.id as a string but it somehow become a number, so the compare couldn't match since it's number vs string. So i put everything as a string to be simplier
- Well we need to have BTT accessible to the old owner but just remove the possibility to add BTT new token (the comment from the list remove it from the whitelist of possible added token)
- I have no idea. Script generate himself from trongrid API. So I guess it's should be expected ?

script/crypto-assets-importer/tron/trc10-tokens.js
henrily-ledger: I'll remove it

ghost (approved):

PR #844, opened 2022-04-27
Fix memo serialization.

matthewcarlreetz:
Fixes incorrect serialization of memo field which was causing transaction verification to fail.


allenan (approved):
thanks, this should fix the issue with burn txn serialization

0ex-d (approved):

PR #866, opened 2022-05-17
LIVE-1810 - ethersproject upgrade

alexalouit:
Upgrade ethersproject and do transaction test


haammar-ledger:
Replaced with https://github.com/LedgerHQ/ledger-live/pull/136

